PR #2, opened 2021-07-23
8.0 skill sharpening project

Ytahirioct82:
finished the skill project functions.

### DIFF
--- a/src/06-loops.js
+++ b/src/06-loops.js
@@ -11,7 +11,14 @@
  *  //> 25
  *  (3 + 4 + 5 + 6 + 7 = 25)
  */
-function rangeSum(min, max) {}
+function rangeSum(min, max) {
+  let result = 0;
+  for (let i = min; i <= max; i++) {
+    result = result + i;
+  }
+
+  return result;
+}
 
 /**
  * rangeOdd()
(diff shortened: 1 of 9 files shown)


src/06-loops.js
```diff
@@ -61,7 +84,16 @@ function rangeEveryOther(min, max) {}
...
+function containsWhileLoop(array, target) {
+  let i = 0;
+
+  while (i < array.length) {
+    i++;
+    if (array.indexOf(target) !== -1) {
+      return true;
+    } else return false;
+  }
+}
```
abbreviatedman, line 96: You pass the test (solve the problem with a while loop), but `.indexOf` is actually doing all the work. Your function could be simply `return array.indexOf(target) !== -1` and you'd return the right result! Your `while` loop is jut for show here, but see if you can USE it to solve the problem WITHOUT `.indexOf` or `.includes`.

abbreviatedman: And the same for the `for` loop contains problem below.

src/06-loops.js
```diff
@@ -117,7 +163,15 @@ function targetCount(array, target) {}
...
-function firstIndexFound(array, target) {}
+function firstIndexFound(array, target) {
+  let result = -1;
+  for (let i = 0; i < array.length; i++) {
+    if (array[i] === target) {
+      return (result = i);
+    }
+  }
+  return result;
+}
```
abbreviatedman, line 174: You don't need the `result =` part in your return; `result = i` just evaluates to `i`, so you could simply write `return i`.

PR #11, opened 2021-07-26
8.0 skill sharpening

DylanLovell93:
Enjoyed the review!

### DIFF
--- a/src/02-data-types.js
+++ b/src/02-data-types.js
@@ -9,7 +9,10 @@
  *  getLengthOfString("down");
  *  //> 4
  */
-function getLengthOfString(str) {}
+function getLengthOfString(str) {
+  //use .length to get the length of the string and return it
+  return str.length;
+}
 
 /**
  * convertToNumber()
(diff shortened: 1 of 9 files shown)


src/02-data-types.js
```diff
@@ -22,7 +25,14 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  //create result value to store result
+  let result = 0;
+  //Convert the string to a number. If successful, our result is the converted number, else give an error
+  typeof Number(val) === "number" ? result = Number(val) : result = "Input is not a number"
```
abbreviatedman, line 32: You don't want the second and third parts of a ternary to be a statement, just an expression to evaluate to. Then you do something with whichever value it evaluates to. So, for the above, you could change it to:

```javascript
result = typeof Number(val) === "number" ? Number(val) : "Input is not a number"
```

PR #32, opened 2021-07-28
add solutions the javascript files in src folder

RohanPursuit:


src/03-control-flow.js
```diff
@@ -179,7 +262,12 @@ function evenOrOdd(a) {}
  *  evenOrOddWithTernary(8);
  *  //> "Even"
  */
-function evenOrOddWithTernary(a) {}
+function evenOrOddWithTernary(a) {
+  let evenOrOdd = null
+  a % 2 === 0 ? evenOrOdd = 'Even' : evenOrOdd = 'Odd'
```
abbreviatedman, line 267: `a % 2 === 0 ? evenOrOdd = 'Even' : evenOrOdd = 'Odd'`
can be rewritten as:

`evenOrOdd = a % 2 === 0 ? 'Even' : 'Odd'`

You generally don't want to have a statement within your ternary, just a pair of values that you want the whole thing to evaluate to. Then you do something with that value. (In this case, assign it to `evenOrOdd`. Or even return it without putting it in a variable!)

PR #36, opened 2021-07-29
Problem Solutions 

angie3j:


src/06-loops.js
```diff
@@ -11,8 +11,16 @@
  *  //> 25
  *  (3 + 4 + 5 + 6 + 7 = 25)
  */
-function rangeSum(min, max) {}
-
+function rangeSum(min, max) {
```
Angiepmunoz, line 14: the test file for loops is not running. I think there could be an infinite loop somewhere, but I couldnt find where.

PR #43, opened 2021-07-29
Skills sharpening project

imolina212:


src/09-word-problems.js
```diff
@@ -127,8 +178,57 @@ function compareLocations(address1, address2) {}
...
+//   else {
+//     assignments.status =
+//       "SCORE: " + assignments.score.received / assignments.score.max;
+//   }
 
+//   return assignments;
+// }
+function gradeAssignments(assignments) {
+  for (let assign of assignments) {
+    if (assign.kind === "PASS-FAIL") {
```
abbreviatedman, line 213: A compound boolean like this:

```javascript
(assign.kind === 'PASS-FAIL' && assign.score.received === assign.score.max)
```

is usually a lot more readable than these ifs inside each other ("nested" ifs).

PR #78, opened 2022-01-28
Completed-skill-sharpening-lab

NikeshW:


src/01-functions.js
```diff
@@ -24,7 +26,9 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet(name) {
+  return ("Hello, " + name + ".");
```
myrasmith, line 30: style comment: the parens here aren't necessary

src/01-functions.js
```diff
@@ -112,7 +124,13 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+  if (typeof a === typeof b) {
```
myrasmith, line 128: Is the conditional here necessary? Try refactoring without it. (Hint: your solution may be a single return statement)

src/06-loops.js
```diff
@@ -83,7 +107,13 @@ function containsWhileLoop(array, target) {}
  *  containsForLoop([ "left", "up", "right" ], "down");
  *  //> false
  */
-function containsForLoop(array, target) {}
+  function containsForLoop(array, target) {
+    for(let n = 0; n <= array.length; n++)
```
myrasmith, line 111: This for loop is missing curly braces. The code happens works since there is only a single statement inside the for loop, but you should always use curly braces to denote the start and end of the body of a loop.

src/09-word-problems.js
```diff
@@ -127,7 +162,33 @@ function compareLocations(address1, address2) {}
     //>   },
     //> ];
  */
-function gradeAssignments(assignments) {}
+function gradeAssignments(assignments) {
```
myrasmith, line 165: Nice work on this function!

It's got some complicated nested logic, so I think you could benefit by splitting up your work into smaller sub problems using helper functions. Could you try refactoring?

PR #82, opened 2022-01-28
Still working on it

jumanji16y:


myrasmith:
I don't see your work on the loops problems - did you commit your work on them?

src/07-even-more-loops.js
```diff
@@ -44,7 +58,15 @@ function shoutWhileLoop(array) {}
  *  shoutForOfLoop([ "A", "Very", "Happy", "Array" ]);
  *  //> [ "A!", "Very!", "Happy!", "Array!" ];
  */
-function shoutForOfLoop(array) {}
+function shoutForOfLoop(array) {
+  let arr = [];
+  let i = 0;
+  for (arr of array) {
```
myrasmith, line 64: Always use let or const to declare the variable in a for...of loop!

src/07-even-more-loops.js
```diff
@@ -57,7 +79,15 @@ function shoutForOfLoop(array) {}
  *  sumArray([ 10, 0, 10, 11 ]);
  *  //> 31
  */
-function sumArray(array) {}
+function sumArray(array) {
+  //let total = 0;
+  let arr = 0;
```
myrasmith, line 84: This is a misleading variable name since the value the variable holds is a number, not an array. Could you pick a name that better represents what the variable is used for?

src/09-word-problems.js
```diff
@@ -40,7 +51,13 @@ function applyDiscount(priceInCents, age, hasMembership) {}
     getCartTotal(cart);
  *  //> "$30.00"
  */
-function getCartTotal(products) {}
+function getCartTotal(products) {
+  let total = 0;
+  if (products.name) {
```
myrasmith, line 56: What is `products`? Does it have a `.name` property? Is it an object? An array of objects?

jumanji16y:
> I don't see your work on the loops problems - did you commit your work on them?

Yes sorry I just noticed I didn't commit anything for this file.  I will write an email explaining in more detail later.  I will commit what I have now.

PR #97, opened 2022-01-29
lab completed

DavidMendoza97:


src/01-functions.js
```diff
@@ -39,7 +43,10 @@ function greet() {}
  *  add(10, 20);
  *  //> 30
  */
-function add() {}
+function add(a, b) {
+  let sum = a + b;
```
gigimscarborough, line 47: we could have saved a variable here and returned just a+b

src/01-functions.js
```diff
@@ -54,7 +61,10 @@ function add() {}
  *  multiply(2, 5);
  *  //> 10
  */
-function multiply() {}
+function multiply(a, b) {
+  let sum = a * b;
```
gigimscarborough, line 65: here too! no variable necessary :)

src/01-functions.js
```diff
@@ -74,7 +84,10 @@ function multiply() {}
  *  average(10, 6);
  *  //> 8 // 10 + 6 = 16; 16/2 = 8;
  */
-function average() {}
+function average(a, b) {
+  let sum = (a + b) / 2;
```
gigimscarborough, line 88: this variable represents the average not the sum now so I would either change the variable to average or just return the average.

src/01-functions.js
```diff
@@ -93,7 +106,10 @@ function average() {}
  *  averageThree(10, 11, 19);
  *  //> 10 // 10 + 11 + 19 = 30; 30 / 3 = 10;
  */
-function averageThree() {}
+function averageThree(a, b, c) {
+  let sum = (a + b + c) / 3;
```
gigimscarborough, line 110: Here too! Rename the variable or let's just return the avg.

src/01-functions.js
```diff
@@ -112,7 +128,13 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+  if (typeof a === typeof b) {
```
gigimscarborough, line 132: This will work for this problem, but if you want to give yourself a little bonus I want to challenge you to think of some cases where this would not work and how you would handle that programmatically!

src/01-functions.js
```diff
@@ -131,7 +153,13 @@ function compareTypes() {}
  *  flipSign(50);
  *  //> -50
  */
-function flipSign() {}
+function flipSign(a) {
+  if (a === Number(-a)) {
```
gigimscarborough, line 157: This works but consider is your intention clear to other developers. Maybe there are more straight forward ways we could we could make this sign flip. Hint: multiplying by a negative.

src/01-functions.js
```diff
@@ -168,7 +200,15 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+  let str = a.split('');
```
gigimscarborough, line 204: Does this variable accurately describe the data type of the value it holds?

src/01-functions.js
```diff
@@ -168,7 +200,15 @@ function joinStringsWithSpaces() {}
...
+  let newStr = [];
```
gigimscarborough, line 205: this one too. Is it a str?

src/02-data-types.js
```diff
@@ -79,7 +97,13 @@ function convertToWhisperText(text) {}
  *  checkIfCharacterIsInString("hello there", "a");
  *  //> false
  */
-function checkIfCharacterIsInString(text, character) {}
+function checkIfCharacterIsInString(text, character) {
+  if (text.includes(character)) {
```
gigimscarborough, line 101: Remember that the return value of the .includes method is a boolean. We could refactor this function to just return the value of text.includes(character)

src/02-data-types.js
```diff
@@ -95,7 +119,13 @@ function checkIfCharacterIsInString(text, character) {}
  *  isEven(11);
  *  //> false
  */
-function isEven(num) {}
+function isEven(num) {
+  if (num % 2 === 0) {
```
gigimscarborough, line 123: Here too! Remember that comparisons evaluate to booleans so we can just return the comparison.

src/02-data-types.js
```diff
@@ -111,7 +141,13 @@ function isEven(num) {}
  *  isOdd(11);
  *  //> true
  */
-function isOdd(num) {}
+function isOdd(num) {
+  if (num % 2 !== 0) {
```
gigimscarborough, line 145: Here too! Just return the comparison.

src/02-data-types.js
```diff
@@ -127,7 +163,13 @@ function isOdd(num) {}
  *  isTruthy(null);
  *  //> false
  */
-function isTruthy(val) {}
+function isTruthy(val) {
+  if (val) {
```
gigimscarborough, line 167: Remember we can use the !! operator here to check the if a value is truthy and simply return it, i.e. return !!val

src/02-data-types.js
```diff
@@ -143,7 +185,13 @@ function isTruthy(val) {}
  *  isFalsy(" ");
  *  //> false
  */
-function isFalsy(val) {}
+function isFalsy(val) {
+  if (!val){
```
gigimscarborough, line 189: here we can just return !val since it will evaluate to a boolean.

src/03-control-flow.js
```diff
@@ -6,15 +6,21 @@
...
  * EXAMPLE:
  *  isEqual(10, 10);
  *  //> true
  *
  *  isEqual(10, "10");
  *  //> false
  */
-function isEqual(a, b) {}
+function isEqual(a, b) {
+  if (a === b) {
```
gigimscarborough, line 18: Just return the comparison (a === b)

src/03-control-flow.js
```diff
@@ -86,7 +114,15 @@ function positiveNegativeOrZero(a) {}
  *  aroundTheWorldGreeting();
  *  //> "Hello World"
  */
-function aroundTheWorldGreeting(language) {}
+function aroundTheWorldGreeting(language = 'english') {
```
gigimscarborough, line 117: nice use of a default value!

src/03-control-flow.js
```diff
@@ -128,7 +173,19 @@ function aroundTheWorldGreetingWithSwitch(language) {}
  *  calculateLetterGrade(99);
  *  //> "A"
  */
-function calculateLetterGrade(grade) {}
+function calculateLetterGrade(grade) {
```
gigimscarborough, line 176: this might have lent itself better to using a switch.

src/04-arrays.js
```diff
@@ -5,12 +5,16 @@
...
  * @returns {*[]}
- *
+ * 
  * EXAMPLE:
  *  createEmptyArray();
  *  //> []
  */
-function createEmptyArray() {}
+function createEmptyArray() {
+  let arr = [];
```
gigimscarborough, line 14: no need for variable, simply return empty arr.

src/04-arrays.js
```diff
@@ -135,7 +163,11 @@ function removeElementFromFrontOfArray(array) {}
  *  getMiddleElement([ 10, null, "30" ]);
  *  //> null
  */
-function getMiddleElement(array) {}
+function getMiddleElement(array) {
+let mid = array[Math.floor((array.length - 1) / 2)];
```
gigimscarborough, line 167: Nice check with the Math.floor but considering the arrays are always odd, do we need it?

src/05-objects.js
```diff
@@ -6,10 +6,14 @@
...
  * EXAMPLE:
- *  createEmptyObject()
+ *  createEmptyObject() 
  *  //> {}
  */
-function createEmptyObject() {}
+function createEmptyObject() {
+  let obj = {};
+
+  return obj;
```
gigimscarborough, line 15: we could just return an object literal here.

src/05-objects.js
```diff
@@ -35,7 +44,12 @@ function createObjectWithValue(val) {}
  *  createObjectWithKey("left");
  *  //> { left: false }
  */
-function createObjectWithKey(key) {}
+function createObjectWithKey(key) {
+  let obj = {};
+  obj[key] = {};
```
gigimscarborough, line 49: i know we could set this equal to anything, but probably more appropriate in this case to set the value as null.

src/06-loops.js
```diff
@@ -137,7 +193,14 @@ function firstIndexFound(array, target) {}
  *  lastIndexFound([ "left", "right", "left" ], "up");
  *  //> -1
  */
-function lastIndexFound(array, target) {}
+function lastIndexFound(array, target) {
+  for (let i = array.length; i > 0; i--) {
```
gigimscarborough, line 197: With better testing, this code should not pass. For this loop to accurately do what we intend, i should equal array.length-1 and our while condition should be i >= 0. As its stands now if the last element is first in the array we wont accurately return an index.

src/06-loops.js
```diff
@@ -165,7 +235,16 @@ function timesIndex(array) {}
...
+function cumulativeSum(array) {
+  let arr = [];
+  let num = 0;
+  
+  for(let i = 0; i < array.length; i++) {
+    num += array[i];
+    arr.push(num);     
+  }
+  return arr;
+}
```
gigimscarborough, line 248: nice solution here!

src/07-even-more-loops.js
```diff
@@ -126,7 +191,13 @@ function findLargest(array) {}
  *  findEqual([ 0, 11, -2, 5 ], 9);
  *  //> false
  */
-function findEqual(array, selected) {}
+function findEqual(array, selected) {
+  if (array.includes(selected)) {
```
gigimscarborough, line 195: Return just array.includes(selected)

src/09-word-problems.js
```diff
@@ -11,15 +11,31 @@
...
  *  applyDiscount(1000, 9, true);
  *  //> 700
  */
-function applyDiscount(priceInCents, age, hasMembership) {}
+function applyDiscount(priceInCents, age, hasMembership) {
+  let result =0
+
+  if (hasMembership && age >= 65 ){
+    result = priceInCents * .70;
+  } else if (hasMembership &&  age <= 10 ){
```
gigimscarborough, line 27: consider combining this condition with our first condition.

src/09-word-problems.js
```diff
@@ -80,7 +103,19 @@ function getCartTotal(products) {}
     compareLocations(address1, address2);
     //> "Same city."
  */
-function compareLocations(address1, address2) {}
+function compareLocations(address1, address2) {
```
gigimscarborough, line 106: consider using a switch here... these conditions are super long.

gigimscarborough (commented):
Left a few comments of things we could change. One function that should fully be rewritten but overall great work!!! Do a dance, you should be really pleased!

PR #102, opened 2022-01-30
Skills Sharpening Project

sbsimonsays:


src/01-functions.js
```diff
@@ -74,7 +82,9 @@ function multiply() {}
  *  average(10, 6);
  *  //> 8 // 10 + 6 = 16; 16/2 = 8;
  */
-function average() {}
+function average(a, b) {
+  return ((a+b)/2);
```
mckiernantim, line 86: I would space out the operations happening inside your parenthesis on 86.  Also we dont need to wrap the entire operation in parenthesis as `a` and `b` will be added together first ( they're in parenthesis ) then the division will occur.

src/01-functions.js
```diff
@@ -93,7 +103,9 @@ function average() {}
  *  averageThree(10, 11, 19);
  *  //> 10 // 10 + 11 + 19 = 30; 30 / 3 = 10;
  */
-function averageThree() {}
+function averageThree(a, b, c) {
+  return ((a+b+c)/3);
```
mckiernantim, line 107: Same as the above note

src/02-data-types.js
```diff
@@ -22,7 +24,9 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  return Number(parseInt(val));
```
mckiernantim, line 28: Be careful with calling `parseInt()` with numbers.  What would happen if we passed in `12.5` to our function?  Simply calling a `Number` constructor is enough to insure we have a number returned from the function.

src/02-data-types.js
```diff
@@ -36,7 +40,9 @@ function convertToNumber(val) {}
  *  convertToString(99);
  *  //> "99"
  */
-function convertToString(val) {}
+function convertToString(val) {
+  return val.toString();
```
mckiernantim, line 44: `toString()` definitely works in the case but remember we can also simply write : `val +""` to convert our val implicitly to a string

src/03-control-flow.js
```diff
@@ -28,7 +35,13 @@ function isEqual(a, b) {}
  *  findLarger(19, 7);
  *  //> 19
  */
-function findLarger(a, b) {}
+function findLarger(a, b) {
```
mckiernantim, line 38: remember that here we know that if `a > b` we will return true.  Because of this we actually don't need an `else` block at all as the only time we will reach the else block is if `b >= a`.

src/03-control-flow.js
```diff
@@ -86,7 +119,15 @@ function positiveNegativeOrZero(a) {}
  *  aroundTheWorldGreeting();
  *  //> "Hello World"
  */
-function aroundTheWorldGreeting(language) {}
+function aroundTheWorldGreeting(language) {
```
mckiernantim, line 122: Lets keep our indentation proper here.  Your function body should all be indented once more over.

mckiernantim (commented):
Great work here Spencer.  Really clean code, well organized, expressive and all our tests are passing.  The only thing I think we need to be mindful of is our spacing and indentation and semi colon placement.  This is totally expected for a new developer but the sooner we can begin to practice these habits the sooner they become second nature.  Keep an eye on spacing around operators and function expression / invocation and make sure that we have semi colons at the end of our statements.

PR #104, opened 2022-01-30
finished 8-0 skill sharpening project!

sfaber02:


src/01-functions.js
```diff
@@ -131,7 +145,9 @@ function compareTypes() {}
  *  flipSign(50);
  *  //> -50
  */
-function flipSign() {}
+function flipSign(a) {
+  return 0 - a;
```
mckiernantim, line 149: We could shorten this to simply `-a` if we wished

src/01-functions.js
```diff
@@ -74,7 +82,9 @@ function multiply() {}
  *  average(10, 6);
  *  //> 8 // 10 + 6 = 16; 16/2 = 8;
  */
-function average() {}
+function average(a, b) {
+  return ((a+b) / 2);
```
mckiernantim, line 86: I would space out the arithmetic assignment to make it slightly easier to read.  Also we don't need the surrounding `()` in this case.

src/02-data-types.js
```diff
@@ -127,7 +143,10 @@ function isOdd(num) {}
  *  isTruthy(null);
  *  //> false
  */
-function isTruthy(val) {}
+function isTruthy(val) {
+  // return val ? true : false;
```
mckiernantim, line 147: Moving forward we should try to remove all of our commented out code before pushing.  Not a huge deal but a good habit to get into as we work.

src/03-control-flow.js
```diff
@@ -86,7 +94,16 @@ function positiveNegativeOrZero(a) {}
  *  aroundTheWorldGreeting();
  *  //> "Hello World"
  */
-function aroundTheWorldGreeting(language) {}
+function aroundTheWorldGreeting(language = 'english') {
```
mckiernantim, line 97: We could space out the ` else if ` from the code blocks to make it a bit cleaner

src/03-control-flow.js
```diff
@@ -109,7 +126,21 @@ function aroundTheWorldGreeting(language) {}
  *  aroundTheWorldGreetingWithSwitch();
  *  //> "Hello World"
  */
-function aroundTheWorldGreetingWithSwitch(language) {}
+function aroundTheWorldGreetingWithSwitch(language) {
+  //i used a switch for the above problem before i saw this one
```
mckiernantim, line 130: Love the comment - I did the same thing when I did this

src/03-control-flow.js
```diff
@@ -128,7 +159,25 @@ function aroundTheWorldGreetingWithSwitch(language) {}
  *  calculateLetterGrade(99);
  *  //> "A"
  */
-function calculateLetterGrade(grade) {}
+function calculateLetterGrade(grade) {
```
mckiernantim, line 162: Remember we don't need a `break` statement if we have a `return` in the switch.  The return will stop the execution of the case and return the value we specify.

src/05-objects.js
```diff
@@ -103,7 +118,10 @@ function addValueToObject(object, val) {}
  *  addKeyToObject({ right: false }, "left");
  *  //> { right: false, left: true }
  */
-function addKeyToObject(object, key) {}
+function addKeyToObject(object, key) {
+  object[key] = '420 hell yeah';
```
mckiernantim, line 122: While currently legal in the state of New York ( and should be everywhere ) it is worth remembering that references to cannabis and its related cultures may rub some recruiters the wrong way.  While highly, highly unlikely it is possible that someone could read this PR at some point.  I would suggest simply being aware of this fact before removing you having fun while coding :)

sfaber02:
Lol good point Tim. Gotta keep it professional. Hopefully I'll have a lot
more interesting projects for recruiters to look at by that point and
they'll never find out about my transgressions 😬.

src/06-loops.js
```diff
@@ -117,7 +159,16 @@ function targetCount(array, target) {}
  *  firstIndexFound([ "left", "right", "left" ], "up");
  *  //> -1
  */
-function firstIndexFound(array, target) {}
+function firstIndexFound(array, target) {
+  let foundIndex = -1;
+  for(let i = 0; i < array.length; i++){
+    if (array[i] === target){
+      foundIndex = i;
```
mckiernantim, line 166: We could shorten our code here to simply return `i` if we find the target.  This saves us from needing to write a break statement

src/06-loops.js
```diff
@@ -137,7 +188,16 @@ function firstIndexFound(array, target) {}
...
  *  //> -1
  */
-function lastIndexFound(array, target) {}
+function lastIndexFound(array, target) {
+  let foundIndex = -1;
+  for(let i = array.length; i >= 0; i--){
+    if (array[i] == target){
+      foundIndex = i;
+    }
+    if (foundIndex != -1) break;
```
mckiernantim, line 197: Here we can move our `break` statement inside of our `if` block.  Since we are iterating backwards and have found an match there is nothing wrong with simply breaking out of loop early as soon as we reassign `foundIndex`

src/06-loops.js
```diff
@@ -151,7 +211,12 @@ function lastIndexFound(array, target) {}
  *  //> [ 7 * 0, 10 * 1, 11 * 2 ]
  *  //> [ 0, 10, 22 ]
  */
-function timesIndex(array) {}
+function timesIndex(array) {
+  array.forEach((e,i) => {
```
mckiernantim, line 215: nice `forEach` - just make sure you know that the  what each argument is actually doing under the hood

src/07-even-more-loops.js
```diff
@@ -28,7 +34,14 @@ function shoutForLoop(array) {}
  *  shoutWhileLoop([ "A", "Very", "Happy", "Array" ]);
  *  //> [ "A!", "Very!", "Happy!", "Array!" ];
  */
-function shoutWhileLoop(array) {}
+function shoutWhileLoop(array) {
+  let newArray = [];
+  while (array.length > 0){
+    //console.log (array.splice(0,1));
```
mckiernantim, line 40: Start thinking about removing console.logs before pushing - good habit to get into in case engineers from teams go snooping around your github

src/07-even-more-loops.js
```diff
@@ -57,7 +76,10 @@ function shoutForOfLoop(array) {}
  *  sumArray([ 10, 0, 10, 11 ]);
  *  //> 31
  */
-function sumArray(array) {}
+function sumArray(array) {
+  if (!array.length) return 0;
+  return array.reduce((a, b) => a + b);
```
mckiernantim, line 81: Great use of `reduce` like other higher order functions just be clear that we know what exactly is happening at each step.

src/07-even-more-loops.js
```diff
@@ -96,7 +134,9 @@ function evenArray(array) {}
  *  findSmallest([ 0, 11, -2, 5 ]);
  *  //> -2
  */
-function findSmallest(array) {}
+function findSmallest(array) {
+  return array.sort()[0];
```
mckiernantim, line 138: `.sort()` has some interesting implementations.  Could you alter this function to sort our array the other direction?

src/07-even-more-loops.js
```diff
@@ -126,7 +168,9 @@ function findLargest(array) {}
  *  findEqual([ 0, 11, -2, 5 ], 9);
  *  //> false
  */
-function findEqual(array, selected) {}
+function findEqual(array, selected) {
+  return array.includes(selected);
```
mckiernantim, line 172: how would we write this without includes?  It's always a good idea to see if we can code up our own version of native methods.

mckiernantim (commented):
Shawn.  Excellent work.  Keep pushing and start taking a look at HTML / CSS.  Our trajectory should be HTML / CSS => React.  I think we can get there sooner than you think.  Other than that we had a couple of redundant breaks, some console.log statements to remove and thats about it.  Nice.

src/05-objects.js
mckiernantim: I'm hoping the humor came through on your reading of this comment.  Fantastic use of the word 'transgressions'

sfaber02: oh yeah 100% I laughed when you pointed it out.  I do see your point though!

src/07-even-more-loops.js
sfaber02: Funnily enough I spent longer than I want to admit writing my own alphabetical sort for a lvl 8 codewars problem.  Maybe a day later I found out about .sort(). Some expletives may have been said when I made that discovery.

PR #105, opened 2022-01-30
Skill Sharpening Project!

zaneahmed1:


src/01-functions.js
```diff
@@ -112,7 +124,13 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+  if (typeof a === typeof b){
```
gigimscarborough, line 128: Remember that this comparison evaluates to a boolean so we can just return it. Also, this will work for this problem, but if you want to give yourself a little bonus I want to challenge you to think of some cases where this would not work and how you would refactor to handle that programmatically!

src/01-functions.js
```diff
@@ -168,7 +190,9 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+  return a.charAt(0) + a.charAt(a.length-1)
```
gigimscarborough, line 194: We could do this in a more straightforward way by simply using bracket notation.

src/02-data-types.js
```diff
@@ -79,7 +89,13 @@ function convertToWhisperText(text) {}
  *  checkIfCharacterIsInString("hello there", "a");
  *  //> false
  */
-function checkIfCharacterIsInString(text, character) {}
+function checkIfCharacterIsInString(text, character) {
+  if(text.includes(character) === true){
```
gigimscarborough, line 93: Here again we can just return this comparison.

src/02-data-types.js
```diff
@@ -95,7 +111,13 @@ function checkIfCharacterIsInString(text, character) {}
  *  isEven(11);
  *  //> false
  */
-function isEven(num) {}
+function isEven(num) {
+  if(num % 2 === 0){
```
gigimscarborough, line 115: Return comparison only here too!

src/02-data-types.js
```diff
@@ -111,7 +133,13 @@ function isEven(num) {}
  *  isOdd(11);
  *  //> true
  */
-function isOdd(num) {}
+function isOdd(num) {
+  if(num % 2 !== 0){
```
gigimscarborough, line 137: Return comparison only here too!

src/02-data-types.js
```diff
@@ -127,7 +155,13 @@ function isOdd(num) {}
  *  isTruthy(null);
  *  //> false
  */
-function isTruthy(val) {}
+function isTruthy(val) {
+  if(val){
```
gigimscarborough, line 159: Remember we can use the `!!` operator here to check the if a value is truthy and simply return it, i.e. return `!!val`

src/02-data-types.js
```diff
@@ -143,7 +177,13 @@ function isTruthy(val) {}
  *  isFalsy(" ");
  *  //> false
  */
-function isFalsy(val) {}
+function isFalsy(val) {
+  if(!val){
```
gigimscarborough, line 181: Here we can just return `!val` since it will evaluate to a boolean.

src/03-control-flow.js
```diff
@@ -14,7 +14,13 @@
  *  isEqual(10, "10");
  *  //> false
  */
-function isEqual(a, b) {}
+function isEqual(a, b) {
+  if(a === b){
```
gigimscarborough, line 18: Return comparison only here!

src/03-control-flow.js
```diff
@@ -86,7 +114,15 @@ function positiveNegativeOrZero(a) {}
  *  aroundTheWorldGreeting();
  *  //> "Hello World"
  */
-function aroundTheWorldGreeting(language) {}
+function aroundTheWorldGreeting(language) {
```
gigimscarborough, line 117: We can give our function a default value here: `function aroundTheWorldGreeting(language="english")`

src/04-arrays.js
```diff
@@ -135,7 +155,9 @@ function removeElementFromFrontOfArray(array) {}
  *  getMiddleElement([ 10, null, "30" ]);
  *  //> null
  */
-function getMiddleElement(array) {}
+function getMiddleElement(array) {
+  return array[((array.length-1)/2)]
```
gigimscarborough, line 159: We don't need the outer `()` inside our brackets

src/06-loops.js
```diff
@@ -151,7 +205,15 @@ function lastIndexFound(array, target) {}
  *  //> [ 7 * 0, 10 * 1, 11 * 2 ]
  *  //> [ 0, 10, 22 ]
  */
-function timesIndex(array) {}
+function timesIndex(array) {
+  arr = []
+  product = 0
```
gigimscarborough, line 210: Oh noes!!! We're creating some global variables here!

src/07-even-more-loops.js
```diff
@@ -57,7 +77,13 @@ function shoutForOfLoop(array) {}
  *  sumArray([ 10, 0, 10, 11 ]);
  *  //> 31
  */
-function sumArray(array) {}
+function sumArray(array) {
+  sum = 0
```
gigimscarborough, line 81: Another 🌍 variable !

src/07-even-more-loops.js
```diff
@@ -28,7 +34,15 @@ function shoutForLoop(array) {}
  *  shoutWhileLoop([ "A", "Very", "Happy", "Array" ]);
  *  //> [ "A!", "Very!", "Happy!", "Array!" ];
  */
-function shoutWhileLoop(array) {}
+function shoutWhileLoop(array) {
+  let arr = []
+  i = 0
```
gigimscarborough, line 39: Another 🌎 variable!

src/06-loops.js
```diff
@@ -165,7 +227,15 @@ function timesIndex(array) {}
  *  //> [ 5, 5 + 2, 5 + 2 + 9 ]
  *  //> [ 5, 7, 16 ]
  */
-function cumulativeSum(array) {}
+function cumulativeSum(array) {
+  arr= []
+  sum = 0
```
gigimscarborough, line 232: More 🌍 variables!!!

src/09-word-problems.js
```diff
@@ -152,7 +198,18 @@ function gradeAssignments(assignments) {}
     createLineOrder(people);
     //> [ "Ray Anderson", "America Marsh", "Wade Carson", "Patience Patel" ]
  */
-function createLineOrder(people) {}
+function createLineOrder(people) {
+  members = []
+  nonmembers = []
```
gigimscarborough, line 203: More 🌍 variables!

gigimscarborough (commented):
Really nice work here Zane! There was some really nicely written code here! Some things I would look out forgoing forward is making sure we're using proper indentation and watching out for those global variables. Otherwise, really nice work! 🚀

PR #107, opened 2022-01-31
sharp project

Sama59:


src/01-functions.js
```diff
@@ -9,7 +9,16 @@
  *  alwaysTrue();
  *  //> true
  */
-function alwaysTrue() {}
+function alwaysTrue(a, b) {
```
gigimscarborough, line 12: Whoa! Looks like we're doing more here that the problem called for. There were no parameters required here(`a,b`)  and we simply needed to `return` the boolean `true` here.

src/01-functions.js
```diff
@@ -24,7 +33,10 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet(name) {
+  name = "Reed";
```
gigimscarborough, line 37: Looks like we're hard coding the `name` Reed here instead of using the value passed in the parameters for name. I think if we omitted the re-assignment of name and just added a period to our interpolated string, these test would pass.

src/01-functions.js
```diff
@@ -112,7 +132,16 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
```
gigimscarborough, line 135: Since our if else statement returns true or false, we can just return our comparison (`return a === b`), as it will evaluate to a boolean.

src/01-functions.js
```diff
@@ -131,8 +160,10 @@ function compareTypes() {}
  *  flipSign(50);
  *  //> -50
  */
-function flipSign() {}
-
+function flipSign(value) {
```
gigimscarborough, line 163: Again, we're hard coding a value here. The idea was to take the input value and if it was negative make it positive and vice versa. Unfortunately in this function you will only ever return the number 1 because you are reassigning the input `value` to be `Math.abs(-1)`(`Math` class should be capitalized) and returning `value`.

src/01-functions.js
```diff
@@ -151,7 +182,11 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
+function joinStringsWithSpaces(a, b, c, d, e) {
+  let str = '';
```
gigimscarborough, line 186: No need to assign this variable to an empty string here.

PR #108, opened 2022-01-31
Submitting project

Teraesa:


src/01-functions.js
```diff
@@ -24,7 +26,9 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet(name) {
+  return "Hello, " + name + ".";
```
mckiernantim, line 30: Nice work - is there a way to do this using string interpolation?

src/02-data-types.js
```diff
@@ -127,7 +155,13 @@ function isOdd(num) {}
  *  isTruthy(null);
  *  //> false
  */
-function isTruthy(val) {}
+function isTruthy(val) {
+  if (val) {
```
mckiernantim, line 159: Can you think of a way to solve this using a ternary ?

src/03-control-flow.js
```diff
@@ -14,7 +14,11 @@
  *  isEqual(10, "10");
  *  //> false
  */
-function isEqual(a, b) {}
+function isEqual(a, b) {
+  if (a === b) {
+    return true;
+  } else return false;
```
mckiernantim, line 20: In situations like line 20 while it is not incorrect to write an ` else ` statement we actually do not need one.  In the event that ` a === b ` returns ` true ` the function will end its run.  So it is safe to say that ` if a !== b ` we can simply return ` false` as there are no other options.

src/03-control-flow.js
```diff
@@ -28,7 +32,13 @@ function isEqual(a, b) {}
  *  findLarger(19, 7);
  *  //> 19
  */
-function findLarger(a, b) {}
+function findLarger(a, b) {
```
mckiernantim, line 35: Could we use a ternary here?

src/03-control-flow.js
```diff
@@ -146,7 +207,13 @@ function calculateLetterGrade(grade) {}
  *  animalCounts("pig", 1);
  *  //> "1 pig"
  */
-function animalCounts(animal, numberOfAnimals) {}
+function animalCounts(animal, numberOfAnimals) {
+  if (numberOfAnimals > 1) {
```
mckiernantim, line 211: Once again, beautifully clean code.  If we wanted to write this even shorter could we do so using a template literal?

src/04-arrays.js
```diff
@@ -24,7 +27,10 @@ function createEmptyArray() {}
  *  createArrayWithTwoElements(true, false);
  *  //> [ true, false ]
  */
-function createArrayWithTwoElements(a, b) {}
+function createArrayWithTwoElements(a, b) {
+  let arr = ["A", "B"];
```
mckiernantim, line 31: While this passes the tests  how can we write this so that whatever arguments are passed end up in the returned array?

src/04-arrays.js
```diff
@@ -10,7 +10,10 @@
  *  createEmptyArray();
  *  //> []
  */
-function createEmptyArray() {}
+function createEmptyArray() {
+  let array = [];
```
mckiernantim, line 14: Can we do this in one line?

src/04-arrays.js
```diff
@@ -106,7 +123,9 @@ function removeElementFromEndOfArray(array) {}
  *  addElementToFrontOfArray([ 10 ], 9);
  *  //> [ 9, 10 ]
  */
-function addElementToFrontOfArray(array, element) {}
+function addElementToFrontOfArray(array, element) {
```
mckiernantim, line 126: Let's look at this again.  Remember that ` unshift() ` does add an elemnent to an array but it actually returns the new length of the array and ` toString() ` converts the data to a string.  How can we add to the array using what we have here to return an array with the new element?

src/04-arrays.js
```diff
@@ -119,7 +138,10 @@ function addElementToFrontOfArray(array, element) {}
  *  removeElementFromFrontOfArray([ 10, 9, 8 ]);
  *  //> 10
  */
-function removeElementFromFrontOfArray(array) {}
+function removeElementFromFrontOfArray(array) {
+  array.shift();
```
mckiernantim, line 142: Letss take another look at this.  What does ` .shift() ` return?  Knowing this what do we need to return from the function?  Do we need to index ` array[0] ` ?

src/04-arrays.js
```diff
@@ -135,7 +157,12 @@ function removeElementFromFrontOfArray(array) {}
  *  getMiddleElement([ 10, null, "30" ]);
  *  //> null
  */
-function getMiddleElement(array) {}
+function getMiddleElement(array) {
+  let mid = [array[array.length / 2 - 1]];
```
mckiernantim, line 161: Lets be careful with our ` [] ` here.  array does not need a bracket around it.  Also lets be careful where we are indexing our ` array `.  What happens with this code if the  ` array[array.length/2] ` turns out to be 1.5?  Can we index an array like this?

src/06-loops.js
```diff
@@ -61,7 +84,21 @@ function rangeEveryOther(min, max) {}
  *  containsWhileLoop([ "left", "up", "right" ], "down");
  *  //> false
  */
-function containsWhileLoop(array, target) {}
+//didnt pass
+function containsWhileLoop(array, target) {
```
mckiernantim, line 88: So this code should actually pass our tests.  It works and does what you intended.  However, based on the very specific ask of the tests this code actually fails.  One of our tests specifically turns your entire function body into a string.  It then checks if that string has the word ` for ` in it - which our commented out code does.  This is a very strange edge case that I have flagged for our curriculum team but it does also serve as a good reminder that comments in our code do not always exist without consequence.  We should start to remove our pseudo code before we attempt tests moving forward.

src/07-even-more-loops.js
```diff
@@ -96,7 +150,21 @@ function evenArray(array) {}
  *  findSmallest([ 0, 11, -2, 5 ]);
  *  //> -2
  */
-function findSmallest(array) {}
+function findSmallest(array) {
+  //infinity is the biggest number, we are looking for the smallest
+  let smallest = Infinity;
```
mckiernantim, line 155: This is an interesting concept here but setting smallest to ` Infinity ` is not needed.  Since we are iterating over an array we can set our smallest to the first element in the array and then move forward with our comparison.

src/07-even-more-loops.js
```diff
@@ -109,7 +177,16 @@ function findSmallest(array) {}
  *  findLargest([ 0, 11, -2, 5 ]);
  *  //> 11
  */
-function findLargest(array) {}
+function findLargest(array) {
+  let largest = -Infinity;
```
mckiernantim, line 181: Once again lets set the value to our initial array item and not use ` Infinity `

src/07-even-more-loops.js
```diff
@@ -126,7 +203,14 @@ function findLargest(array) {}
  *  findEqual([ 0, 11, -2, 5 ], 9);
  *  //> false
  */
-function findEqual(array, selected) {}
+function findEqual(array, selected) {
+  //loop
+  for (let i = 0; i < array.length; i++) {
+    if (array[i] === selected) return true;
```
mckiernantim, line 209: I really like this one line if statement - great habit to get into.

src/08-accessing.js
```diff
@@ -116,7 +132,12 @@ function getLastPhoneNumber(person) {}
  *  getFullName(person);
  *  //> "Rachel Eleanor Rojas"
  */
-function getFullName(person) {}
+function getFullName(person) {
+  let firstName = person.names.first;
```
mckiernantim, line 136: Once again nice work - can we do this without using extra variables?

src/08-accessing.js
```diff
@@ -148,7 +173,16 @@ function getCityAndState(person) {}
...
  */
-function getFullAddress(person) {}
+function getFullAddress(person) {
+  let street = person.address.street;
+  let unit = person.address.unit;
+  let city = person.address.city;
+  let state = person.address.state;
+  let zip = person.address.zip;
+  return (
+    street + " " + unit + "," + " " + city + "," + " " + state + "," + " " + zip
```
mckiernantim, line 183: Can we do this with string interpolation?

mckiernantim (commented):
Teraesa we were very close to finishing this project.  Overall you are writing some beautiful code.  Great spacing, great semi colon use,  your indentation is great.  Some things I would recommend are trying to work in ternarys where we can and let's try to shorten our code by only using variables where we need them.  Also - you had a very strange bug that had to do with your comments inside one of your functions.  For the future lets keep our comments outside of our functions just to be 1000% sure that this doesn't happen again. Excellent start!  We will give you time down the road to come back to this.

PR #110, opened 2022-01-31
Skills Sharpened 

jamalRuiz:


src/01-functions.js
```diff
@@ -39,7 +43,9 @@ function greet() {}
  *  add(10, 20);
  *  //> 30
  */
-function add() {}
+function add(a,b) {
+  return a+b
```
mckiernantim, line 47: This code works great but when working with arithmetic operators lets space them out.  Same with out function parameters `(a,b)` => ` (a, b)` and  ` a+b ` => ` a + b `

src/01-functions.js
```diff
@@ -112,7 +124,14 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a,b) {
+  if (typeof(a)===typeof(b)){
```
mckiernantim, line 128: Let's watch our indentation here. - in the same way we have our code spaced out we need to make sure our code is nested properly.  We should indent one tab for each block that we nest in our code.

src/01-functions.js
```diff
@@ -168,7 +191,9 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+  return a.charAt(0) + a.charAt(a.length-1)
```
mckiernantim, line 195: Nice ` charAt() ` call.  Remember we can index strings like arrays.  ` a.charAt(0) ` => `  a[0] `

src/02-data-types.js
```diff
@@ -22,7 +24,9 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  return Number (val)
```
mckiernantim, line 28: When using constructors like `Number` we actually want the ` () ` to not be spaced out. We read this as one call to the `Number` constructor with the arguments passed.

src/02-data-types.js
```diff
@@ -36,7 +40,9 @@ function convertToNumber(val) {}
  *  convertToString(99);
  *  //> "99"
  */
-function convertToString(val) {}
+function convertToString(val) {
+  return val.toString()
```
mckiernantim, line 44: Much like our `charAt()` call we can shorten this to ` val + "" ` and JS will implicitly convert this to a string.

src/02-data-types.js
```diff
@@ -95,7 +109,13 @@ function checkIfCharacterIsInString(text, character) {}
  *  isEven(11);
  *  //> false
  */
-function isEven(num) {}
+function isEven(num) { 
+  if (num%2 === 0)
```
mckiernantim, line 113: Lets make sure that our ` if ` statements have a block after them ` if(thisIsTrue) { } `.  We can do an if statement on one line but if we move to multiple lines we need to add a block.  Also - do we need to put an else statement here or can we simply ` return false ` in the event that we don't return true?

src/03-control-flow.js
```diff
@@ -28,7 +32,13 @@ function isEqual(a, b) {}
  *  findLarger(19, 7);
  *  //> 19
  */
-function findLarger(a, b) {}
+function findLarger(a, b) {
+  if (a>b){
```
mckiernantim, line 36: For if statements lets space them as such ` if(a > b) { } `

src/03-control-flow.js
```diff
@@ -86,8 +118,20 @@ function positiveNegativeOrZero(a) {}
  *  aroundTheWorldGreeting();
  *  //> "Hello World"
  */
-function aroundTheWorldGreeting(language) {}
+function aroundTheWorldGreeting(language) {
+  if (language ==`english`){
+    return "Hello World";
+  }
```
mckiernantim, line 124: The indentation on here is pretty close - if we knock our last  ` else ` statement over one tab we would be good.

src/03-control-flow.js
```diff
@@ -109,7 +153,18 @@ function aroundTheWorldGreeting(language) {}
  *  aroundTheWorldGreetingWithSwitch();
  *  //> "Hello World"
  */
-function aroundTheWorldGreetingWithSwitch(language) {}
+function aroundTheWorldGreetingWithSwitch(language) {
+  switch (language){
```
mckiernantim, line 157: Here lets indent our returns one tab over for each of our cases

src/04-arrays.js
```diff
@@ -92,7 +105,10 @@ function addElementToEndOfArray(array, element) {}
  *  removeElementFromEndOfArray([ 10, 9, 8 ]);
  *  //> 8
  */
-function removeElementFromEndOfArray(array) {}
+function removeElementFromEndOfArray(array) {
+  array=array.pop();
```
mckiernantim, line 109: This technically works but there are some confusing aspects we can smooth out.  First ` .pop() ` already returns whatever item is removed from the array on which it is called.  We could write ` return array.pop() `  but if we do want to save our value in a variable we should choose a different name for our popped off index than ` array `.  Something like ` removed ` or ` index ` is better.

src/05-objects.js
```diff
@@ -22,7 +24,9 @@ function createEmptyObject() {}
  *  createObjectWithValue(19);
  *  //> { whateverKey: 19 }
  */
-function createObjectWithValue(val) {}
+function createObjectWithValue(val) {
```
mckiernantim, line 27: Nice work here.  Make sure we know what the ` key ` in this object is going to evaluate too in addition to its value.

src/05-objects.js
```diff
@@ -118,7 +142,12 @@ function addKeyToObject(object, key) {}
  *  addKeyValuePairToObject({ up: true }, "left", false);
  *  //> { up: true, left: false }
  */
-function addKeyValuePairToObject(object, key, val) {}
+function addKeyValuePairToObject(object, key, val) {
+  // console.log(object,key,val)
```
mckiernantim, line 146: Lets also keep an eye out for our console.log statements.  These won't break our code but we should always remove commented out code or console.log() statements before we push.

src/06-loops.js
```diff
@@ -11,7 +13,15 @@
  *  //> 25
  *  (3 + 4 + 5 + 6 + 7 = 25)
  */
-function rangeSum(min, max) {}
+function rangeSum(min, max) {
+  let sum= 0
+  for (let i= min; i <= max; ++i){
```
mckiernantim, line 18: Lets space out our loops as well.

src/06-loops.js
```diff
@@ -61,7 +86,18 @@ function rangeEveryOther(min, max) {}
  *  containsWhileLoop([ "left", "up", "right" ], "down");
  *  //> false
  */
-function containsWhileLoop(array, target) {}
+function containsWhileLoop(array, target) {
+  let result = false;
+  let i=0;
+  while(i< array.length){
+    if (array[i]=== target){
```
mckiernantim, line 93: Be careful of this extra whitespace as well

src/07-even-more-loops.js
```diff
@@ -96,8 +138,14 @@ function evenArray(array) {}
  *  findSmallest([ 0, 11, -2, 5 ]);
  *  //> -2
  */
-function findSmallest(array) {}
-
+function findSmallest(array) { let smallest = array[0];
```
mckiernantim, line 141: Lets knock this one down a line into the function as well.

src/07-even-more-loops.js
```diff
@@ -109,7 +157,15 @@ function findSmallest(array) {}
  *  findLargest([ 0, 11, -2, 5 ]);
  *  //> 11
  */
-function findLargest(array) {}
+function findLargest(array) {
```
mckiernantim, line 160: This is probably the best indented and formatted function in your project.  Notice how easily we can tell what code is nested where and what our operations are.  All we're missing are a few semicolons at the end of our statements.

src/08-accessing.js
```diff
@@ -148,7 +165,9 @@ function getCityAndState(person) {}
  *  getFullAddress(person);
  *  //> "697 Pine Drive 2A, Staten Island, NY, 10306"
  */
-function getFullAddress(person) {}
+function getFullAddress(person) {
+  return `${person.address.street} ${person.address.unit}, ${person.address.city}, ${person.address.state}, ${person.address.zip}`
```
mckiernantim, line 169: Long strings like this with backticks can be spread across multiple lines.  We should knock this long string down to at least two if not more lines.

src/09-word-problems.js
```diff
@@ -80,7 +100,31 @@ function getCartTotal(products) {}
     compareLocations(address1, address2);
     //> "Same city."
  */
-function compareLocations(address1, address2) {}
+function compareLocations(address1, address2) {
+  if (
+    address1.street === address2.street &&
```
mckiernantim, line 105: Great indentation on this.  We should always aim to set up our code like this.

mckiernantim (commented):
Nice work here! Our biggest issue to keep in mind is our formatting.  Let's make sure that we add semi-colons after all of our statements and we keep our indentation proper.  Other than that our code looks pretty solid.  This is a great place to start out.  If you have any questions on how to style your code Airbnb has an open source style guide that is considered to be the gold standard for JS -->  https://github.com/airbnb/javascript

PR #111, opened 2022-01-31
skill-sharpening-project submission

JimmyAyivor:
completed assignment

### DIFF
--- a/src/01-functions.js
+++ b/src/01-functions.js
@@ -9,7 +9,9 @@
  *  alwaysTrue();
  *  //> true
  */
-function alwaysTrue() {}
+function alwaysTrue() {
+  return true;
+}
 
 /**
  * greet()
@@ -24,7 +26,9 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet(name) {
+  return `Hello, ${name}.`;
+}
 
 /**
  * add()
@@ -39,7 +43,9 @@ function greet() {}
  *  add(10, 20);
  *  //> 30
  */
-function add() {}
+function add(a,b) {
+  return a + b;
+}
 
 /**
  * multiply()
@@ -54,7 +60,9 @@ function add() {}
  *  multiply(2, 5);
  *  //> 10
  */
-function multiply() {}
+function multiply(a,b) {
+  return a * b;
+}
 
 /**
  * average()
@@ -74,7 +82,9 @@ function multiply() {}
  *  average(10, 6);
  *  //> 8 // 10 + 6 = 16; 16/2 = 8;
  */
-function average() {}
+function average(a, b) {
+  return (a + b) / 2;
+}
 
 /**
  * averageThree()
@@ -93,7 +103,9 @@ function average() {}
  *  averageThree(10, 11, 19);
  *  //> 10 // 10 + 11 + 19 = 30; 30 / 3 = 10;
  */
-function averageThree() {}
+function averageThree(a, b, c) {
+  return (a + b + c) / 3;
+}
 
 /**
  * compareTypes()
@@ -112,7 +124,12 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+  if(typeof a === typeof b ){
+   return true;
+  }
+  return false;
+}
 
 /**
  * flipSign()
@@ -131,7 +148,13 @@ function compareTypes() {}
  *  flipSign(50);
  *  //> -50
  */
-function flipSign() {}
+function flipSign(a) {
+  if(a< 0){
+    return Math.abs(a);
+  }
+  return -Math.abs(a);
+  }
+
 
 /**
  * joinStringsWithSpaces()
@@ -151,7 +174,9 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
+function joinStringsWithSpaces(a,b,c,d,e) {
+  return `${a} ${b} ${c} ${d} ${e}`;
+}
 
 /**
  * getFirstAndLastCharacter()
@@ -168,7 +193,10 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+  newString = a[0] + a[a.length-1];
+  return newString;
+}
 
 // Do not change the code below.
 module.exports = {
--- a/src/02-data-types.js
+++ b/src/02-data-types.js
@@ -9,7 +9,9 @@
  *  getLengthOfString("down");
  *  //> 4
  */
-function getLengthOfString(str) {}
+function getLengthOfString(str) {
+  return str.length;
+}
 
 /**
  * convertToNumber()
@@ -22,7 +24,10 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  newVal = Number(val);
+  return newVal;
+}
 
 /**
  * convertToString()
@@ -36,7 +41,10 @@ function convertToNumber(val) {}
  *  convertToString(99);
  *  //> "99"
  */
-function convertToString(val) {}
+function convertToString(val) {
+  newVal = val.toString();
+  return newVal;
+}
 
 /**
  * convertToShoutingText()
@@ -49,7 +57,10 @@ function convertToString(val) {}
  *  convertToShoutingText("Hello There");
  *  //> "HELLO THERE"
  */
-function convertToShoutingText(text) {}
+function convertToShoutingText(text) {
+  newText = text.toUpperCase();
+  return newText;
+}
 
 /**
  * convertToWhisperText()
@@ -62,7 +73,10 @@ function convertToShoutingText(text) {}
  *  convertToWhisperText("Hello There");
  *  //> "hello there"
  */
-function convertToWhisperText(text) {}
+function convertToWhisperText(text) {
+  newText = text.toLowerCase();
+  return newText;
+}
 
 /**
  * checkIfCharacterIsInString()
@@ -79,7 +93,13 @@ function convertToWhisperText(text) {}
  *  checkIfCharacterIsInString("hello there", "a");
  *  //> false
  */
-function checkIfCharacterIsInString(text, character) {}
+function checkIfCharacterIsInString(text, character) {
+  
+  if(text.indexOf(character) == -1){
+    return false;
+  }
+  return true;
+}
 
 /**
  * isEven()
@@ -95,7 +115,12 @@ function checkIfCharacterIsInString(text, character) {}
  *  isEven(11);
  *  //> false
  */
-function isEven(num) {}
+function isEven(num) {
+  if(num % 2 == 0){
+    return true;
+  }
+  return false;
+}
 
 /**
  * isOdd()
@@ -111,7 +136,13 @@ function isEven(num) {}
  *  isOdd(11);
  *  //> true
  */
-function isOdd(num) {}
+function isOdd(num) {
+  if(num % 2 !== 0){
+    return true;
+  }
+  return false;
+
+}
 
 /**
  * isTruthy()
@@ -127,8 +158,12 @@ function isOdd(num) {}
  *  isTruthy(null);
  *  //> false
  */
-function isTruthy(val) {}
-
+function isTruthy(val) {
+  if(val){
+    return true;
+}
+   return false
+}
 /**
  * isFalsy()
  * ---------------------
@@ -143,7 +178,12 @@ function isTruthy(val) {}
  *  isFalsy(" ");
  *  //> false
  */
-function isFalsy(val) {}
+function isFalsy(val) {
+  if(!val){
+    return true;
+}
+   return false
+}
 
 // Do not change the code below.
 module.exports = {
--- a/src/03-control-flow.js
+++ b/src/03-control-flow.js
@@ -14,7 +14,12 @@
  *  isEqual(10, "10");
  *  //> false
  */
-function isEqual(a, b) {}
+function isEqual(a, b) {
+  if(a === b){
+    return true;
+  }
+  return false;
+}
 
 /**
  * findLarger()
@@ -28,7 +33,12 @@ function isEqual(a, b) {}
  *  findLarger(19, 7);
  *  //> 19
  */
-function findLarger(a, b) {}
+function findLarger(a, b) {
+  if(a > b){
+    return a;
+  }
+  return b
+}
 
 /**
  * findLargerOrTie()
@@ -45,7 +55,15 @@ function findLarger(a, b) {}
  *  findLargerOrTie(0, 0);
  *  //> "tie"
  */
-function findLargerOrTie(a, b) {}
+function findLargerOrTie(a, b) {
+  if(a > b){
+    return a;
+  } else if(b > a){
+    return b;
+  } else {
+    return `tie`;
+  }
+}
 
 /**
  * positiveNegativeOrZero()
@@ -65,7 +83,17 @@ function findLargerOrTie(a, b) {}
  *  positiveNegativeOrZero(1);
  *  //> "Positive"
  */
-function positiveNegativeOrZero(a) {}
+function positiveNegativeOrZero(a) {
+  if(a == 0){
+    return `Zero`;
+  }
+  if(a < 0){
+    return `Negative`;
+  }
+  if(a > 0){
+    return `Positive`;
+  }
+}
 
 /**
  * aroundTheWorldGreeting()
@@ -86,7 +114,15 @@ function positiveNegativeOrZero(a) {}
  *  aroundTheWorldGreeting();
  *  //> "Hello World"
  */
-function aroundTheWorldGreeting(language) {}
+function aroundTheWorldGreeting(language = "english") {
+  let lang = language.toLowerCase()
+  if(lang === 'spanish'){
+    return `Hola Mundo`
+  }else if (lang === 'german'){
+    return `Hallo Welt`;
+  }
+  return `Hello World`;
+}
 
 /**
  * aroundTheWorldGreetingWithSwitch()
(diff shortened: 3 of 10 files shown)


JimmyAyivor:
skill-sharpening-project completed!

src/01-functions.js
```diff
@@ -131,7 +148,13 @@ function compareTypes() {}
  *  flipSign(50);
  *  //> -50
  */
-function flipSign() {}
+function flipSign(a) {
```
gigimscarborough, line 151: Nice use of Math.abs here and explicitly asking if the number is less than 0! Super slick! Lets make sure were consistent with our use of white space in our code and indentation though.

src/01-functions.js
```diff
@@ -112,7 +124,12 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+  if(typeof a === typeof b ){
```
gigimscarborough, line 128: This will work for this problem, but if you want to give yourself a little bonus I want to challenge you to think of some cases where this would not work and how you would refactor to handle that programmatically!

src/01-functions.js
```diff
@@ -151,7 +174,9 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
+function joinStringsWithSpaces(a,b,c,d,e) {
+  return `${a} ${b} ${c} ${d} ${e}`;
```
gigimscarborough, line 178: nice use of string interpolation here!

src/02-data-types.js
```diff
@@ -22,7 +24,10 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  newVal = Number(val);
```
gigimscarborough, line 28: Here we could have returned `Number(val)` - no need for a variable we're no using it again in our code!

src/02-data-types.js
```diff
@@ -36,7 +41,10 @@ function convertToNumber(val) {}
  *  convertToString(99);
  *  //> "99"
  */
-function convertToString(val) {}
+function convertToString(val) {
+  newVal = val.toString();
```
gigimscarborough, line 45: Here too! Just return `val.toString()`

src/02-data-types.js
```diff
@@ -49,7 +57,10 @@ function convertToString(val) {}
  *  convertToShoutingText("Hello There");
  *  //> "HELLO THERE"
  */
-function convertToShoutingText(text) {}
+function convertToShoutingText(text) {
+  newText = text.toUpperCase();
```
gigimscarborough, line 61: No variable needed here too!

src/02-data-types.js
```diff
@@ -62,7 +73,10 @@ function convertToShoutingText(text) {}
  *  convertToWhisperText("Hello There");
  *  //> "hello there"
  */
-function convertToWhisperText(text) {}
+function convertToWhisperText(text) {
+  newText = text.toLowerCase();
```
gigimscarborough, line 77: No variable necessary here either

src/02-data-types.js
```diff
@@ -79,7 +93,13 @@ function convertToWhisperText(text) {}
  *  checkIfCharacterIsInString("hello there", "a");
  *  //> false
  */
-function checkIfCharacterIsInString(text, character) {}
+function checkIfCharacterIsInString(text, character) {
+  
+  if(text.indexOf(character) == -1){
```
gigimscarborough, line 98: Remember comparisons evaluate to booleans so we can just `return text.indexOf(character) !== -1`

src/02-data-types.js
```diff
@@ -95,7 +115,12 @@ function checkIfCharacterIsInString(text, character) {}
  *  isEven(11);
  *  //> false
  */
-function isEven(num) {}
+function isEven(num) {
+  if(num % 2 == 0){
```
gigimscarborough, line 119: Here too we can just return the comparison!

src/02-data-types.js
```diff
@@ -111,7 +136,13 @@ function isEven(num) {}
  *  isOdd(11);
  *  //> true
  */
-function isOdd(num) {}
+function isOdd(num) {
+  if(num % 2 !== 0){
```
gigimscarborough, line 140: Return comparison here too!

src/02-data-types.js
```diff
@@ -127,8 +158,12 @@ function isOdd(num) {}
  *  isTruthy(null);
  *  //> false
  */
-function isTruthy(val) {}
-
+function isTruthy(val) {
+  if(val){
```
gigimscarborough, line 162: Remember we can use `!!` operator to check the inherent truthiness of a value. So here we can just return `!!val`.

src/02-data-types.js
```diff
@@ -143,7 +178,12 @@ function isTruthy(val) {}
  *  isFalsy(" ");
  *  //> false
  */
-function isFalsy(val) {}
+function isFalsy(val) {
+  if(!val){
```
gigimscarborough, line 182: Remember `!val` will evaluate to a boolean so we can just return `!val`

src/03-control-flow.js
```diff
@@ -14,7 +14,12 @@
  *  isEqual(10, "10");
  *  //> false
  */
-function isEqual(a, b) {}
+function isEqual(a, b) {
+  if(a === b){
```
gigimscarborough, line 18: We only need to return the comparison here.

src/04-arrays.js
```diff
@@ -24,7 +25,10 @@ function createEmptyArray() {}
  *  createArrayWithTwoElements(true, false);
  *  //> [ true, false ]
  */
-function createArrayWithTwoElements(a, b) {}
+function createArrayWithTwoElements(a, b) {
+newArray = new Array(a,b);
```
gigimscarborough, line 29: Interesting use of the array initializer here! Remember we could have also just put these values in an array literal: `[a, b]`

src/04-arrays.js
```diff
@@ -92,7 +105,10 @@ function addElementToEndOfArray(array, element) {}
  *  removeElementFromEndOfArray([ 10, 9, 8 ]);
  *  //> 8
  */
-function removeElementFromEndOfArray(array) {}
+function removeElementFromEndOfArray(array) {
+  removed = array.pop();
```
gigimscarborough, line 109: No need for this variable. We can `return array.pop()` here.

src/04-arrays.js
```diff
@@ -119,7 +138,10 @@ function addElementToFrontOfArray(array, element) {}
  *  removeElementFromFrontOfArray([ 10, 9, 8 ]);
  *  //> 10
  */
-function removeElementFromFrontOfArray(array) {}
+function removeElementFromFrontOfArray(array) {
+  removed = array.shift();
```
gigimscarborough, line 142: No need for this variable. We can `return array.shift()` here.

src/04-arrays.js
```diff
@@ -135,7 +157,10 @@ function removeElementFromFrontOfArray(array) {}
  *  getMiddleElement([ 10, null, "30" ]);
  *  //> null
  */
-function getMiddleElement(array) {}
+function getMiddleElement(array) {
+  middle = array[Math.floor(array.length / 2)];
```
gigimscarborough, line 161: We can just return this value here!

src/05-objects.js
```diff
@@ -35,7 +39,10 @@ function createObjectWithValue(val) {}
  *  createObjectWithKey("left");
  *  //> { left: false }
  */
-function createObjectWithKey(key) {}
+function createObjectWithKey(key) {
+ 
+  return { [key] : "Banana" };
```
gigimscarborough, line 44: 🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌🍌!

src/05-objects.js
```diff
@@ -103,7 +121,10 @@ function addValueToObject(object, val) {}
  *  addKeyToObject({ right: false }, "left");
  *  //> { right: false, left: true }
  */
-function addKeyToObject(object, key) {}
+function addKeyToObject(object, key) {
+  object[key] = "Oranage";
```
gigimscarborough, line 125: 🍊 you glad I didn't say 🍌?

src/06-loops.js
```diff
@@ -25,7 +31,17 @@ function rangeSum(min, max) {}
...
  *  //> [ 15, 13, 11 ]
  */
-function rangeOdd(min, max) {}
+ function rangeOdd(min, max) {
+  let arr=[];
+  for(let i = max;i>=min;i-- ){
+    if(i % 2 !== 0){
+     arr.push(i);
+    }
+
```
gigimscarborough, line 40: no need for extra line here....

src/06-loops.js
```diff
@@ -25,7 +31,17 @@ function rangeSum(min, max) {}
...
-function rangeOdd(min, max) {}
+ function rangeOdd(min, max) {
+  let arr=[];
+  for(let i = max;i>=min;i-- ){
+    if(i % 2 !== 0){
+     arr.push(i);
+    }
+    
+  }
+
```
gigimscarborough, line 42: or here...

src/06-loops.js
```diff
@@ -39,8 +55,14 @@ function rangeOdd(min, max) {}
  *  rangeEveryOther(11, 18);
  *  //> [ 11, 13, 15, 17 ]
  */
-function rangeEveryOther(min, max) {}
-
+ function rangeEveryOther(min, max) {
+  let arr=[];
+  for(let i = min;i<=max;i++ ){
```
gigimscarborough, line 60: Remember our increment in our for loops doesn't _have_ to be `var++` or `var--`, we can increment however we want. In this case, instead of using `i++` I might have used `i+=2`.

src/06-loops.js
```diff
@@ -39,8 +55,14 @@ function rangeOdd(min, max) {}
...
+      arr.push(i);
+      i += 1
```
gigimscarborough, line 62: Get rid of this here and move to our for loop declaration.

src/06-loops.js
```diff
@@ -61,8 +83,17 @@ function rangeEveryOther(min, max) {}
  *  containsWhileLoop([ "left", "up", "right" ], "down");
  *  //> false
  */
-function containsWhileLoop(array, target) {}
-
+ function containsWhileLoop(array, target) {
+  let i = 0;
+    while (i < array.length) {
+     if(array.indexOf(target) != -1){
```
gigimscarborough, line 89: Here... our while loop isn't really doing the work for us. Instead, by using `array.indexOf(target)` we've outsourced our while loops' work because `.indexOf` does this iteration for us. TLDR- If you took out the `while` loop and just wrote the conditional you would still arrive to the correct outcome because our `while` loop is not really doing anything here. Consider rewriting this to check each element in the array to see if it is our target element...

src/06-loops.js
```diff
@@ -97,7 +134,14 @@ function containsForLoop(array, target) {}
  *  targetCount([ 10, 20, 10, 20, 30 ], 10);
  *  //> 2
  */
-function targetCount(array, target) {}
+function targetCount(array, target) {
+ let count = 0;
+ for(number in array)
```
gigimscarborough, line 139: Remember that since what we care about here is a value in an `Array`, a `for...of` loop would lend itself better here, that way instead of our variable being each index, instead we would have each element.

src/06-loops.js
```diff
@@ -117,7 +161,18 @@ function targetCount(array, target) {}
  *  firstIndexFound([ "left", "right", "left" ], "up");
  *  //> -1
  */
-function firstIndexFound(array, target) {}
+ function firstIndexFound(array, target) {
+  for(let i = 0; i < array.length;i++){
+    if( target ===  array[i]){
+      return i;
+      
+    }else if( target !==  array[i]){
```
gigimscarborough, line 169: we don't need this `else if` here. Remember that by default JavaScript will continue our iteration if our first condition did not return anything.

src/06-loops.js
```diff
@@ -137,7 +192,17 @@ function firstIndexFound(array, target) {}
  *  lastIndexFound([ "left", "right", "left" ], "up");
  *  //> -1
  */
-function lastIndexFound(array, target) {}
+function lastIndexFound(array, target) {
+    for(let i = array.length; i >= 0;i--){
+      if( target ===  array[i]){
+        return i;
+        
+      }else if( target !==  array[i]){
```
gigimscarborough, line 200: Same as above here. We don't need this `else if`, we would continue by default.

src/06-loops.js
```diff
@@ -151,7 +216,12 @@ function lastIndexFound(array, target) {}
  *  //> [ 7 * 0, 10 * 1, 11 * 2 ]
  *  //> [ 0, 10, 22 ]
  */
-function timesIndex(array) {}
+function timesIndex(array) {
```
gigimscarborough, line 219: Remember, this asked for a _new_ `Array` in memory. Here we are returning the same `Array `as our input, but mutated. With a bit stricter testing this likely would not have passed.

src/06-loops.js
```diff
@@ -165,7 +235,12 @@ function timesIndex(array) {}
  *  //> [ 5, 5 + 2, 5 + 2 + 9 ]
  *  //> [ 5, 7, 16 ]
  */
-function cumulativeSum(array) {}
+function cumulativeSum(array) {
```
gigimscarborough, line 238: Same here - not creating a new `Array`

src/07-even-more-loops.js
```diff
@@ -12,7 +12,12 @@
  *  shoutForLoop([ "A", "Very", "Happy", "Array" ]);
  *  //> [ "A!", "Very!", "Happy!", "Array!" ];
  */
-function shoutForLoop(array) {}
+ function shoutForLoop(array) {
```
gigimscarborough, line 15: Mutating here too.

src/07-even-more-loops.js
```diff
@@ -28,7 +33,14 @@ function shoutForLoop(array) {}
  *  shoutWhileLoop([ "A", "Very", "Happy", "Array" ]);
  *  //> [ "A!", "Very!", "Happy!", "Array!" ];
  */
-function shoutWhileLoop(array) {}
+function shoutWhileLoop(array) {
+  let i = 0;
+  while (i < array.length) {
```
gigimscarborough, line 38: Still mutating - now starring `while` loop!

src/07-even-more-loops.js
```diff
@@ -126,8 +185,13 @@ function findLargest(array) {}
  *  findEqual([ 0, 11, -2, 5 ], 9);
  *  //> false
  */
-function findEqual(array, selected) {}
-
+function findEqual(array, selected) {
+  for(let i=0;i< array.length;i++){
+    if(array.includes(selected))
```
gigimscarborough, line 190: Again includes is doing the work here for us no our loop. Refactor this to check each individual element's equality against the selected value.

src/07-even-more-loops.js
```diff
@@ -143,7 +207,11 @@ function findEqual(array, selected) {}
  *  //> [ 1, 11, 2, 3, 4, 9 ]
  */
 
-function removeDuplicates(array) {}
+function removeDuplicates(array) {
+  array  = [...new Set(array)];
```
gigimscarborough, line 211: I have questions for you....... ⁉️

src/08-accessing.js
```diff
@@ -103,7 +114,10 @@ function getFirstPhoneNumber(person) {}
  *  getLastPhoneNumber(person);
  *  //> 7185558611
  */
-function getLastPhoneNumber(person) {}
+function getLastPhoneNumber(person) {
+  lastNumber = person.numbers[person.numbers.length - 1];
```
gigimscarborough, line 118: No variable necessary here.

src/08-accessing.js
```diff
@@ -116,7 +130,10 @@ function getLastPhoneNumber(person) {}
  *  getFullName(person);
  *  //> "Rachel Eleanor Rojas"
  */
-function getFullName(person) {}
+function getFullName(person) {
+fullName = `${person.names.first} ${person.names.middle} ${person.names.last}`
```
gigimscarborough, line 134: No variable necessary here.

src/08-accessing.js
```diff
@@ -132,7 +149,10 @@ function getFullName(person) {}
  *  getCityAndState(person);
  *  //> "Staten Island, NY"
  */
-function getCityAndState(person) {}
+function getCityAndState(person) {
+  address = `${person.address.city}, ${person.address.state}`;
```
gigimscarborough, line 153: No variable necessary here.

src/08-accessing.js
```diff
@@ -171,7 +194,12 @@ function getFullAddress(person) {}
       numbers: [7185550921, 7185558611],
     };
  */
-function getFlatObject(person) {}
+function getFlatObject(person) {
+  const arr = Object.assign(person.names, person.address);
```
gigimscarborough, line 198: We should probably choose a better variable name here. Remember this variable represents an `Object`, not an `Array`. Also, I would have not chosen to use `Object.assign`. Remember this method _mutates_ the first argument passed to it. Now in memory the value of person.names has been manipulated to contain the data in person.address. Remember that this problem called for you to create a _new_ `Object` in memory, not manipulate the input.

src/09-word-problems.js
```diff
@@ -80,7 +101,18 @@ function getCartTotal(products) {}
     compareLocations(address1, address2);
     //> "Same city."
  */
-function compareLocations(address1, address2) {}
+function compareLocations(address1, address2) {
```
gigimscarborough, line 104: Whoa!!!!! These are some gnarly conditionals here. Consider a `switch` statement might lend itself a bit better here.

src/09-word-problems.js
```diff
@@ -152,7 +211,22 @@ function gradeAssignments(assignments) {}
     createLineOrder(people);
     //> [ "Ray Anderson", "America Marsh", "Wade Carson", "Patience Patel" ]
  */
-function createLineOrder(people) {}
+function createLineOrder(people) {
+  let arr1 = []
+  let arr2 = []
```
gigimscarborough, line 216: We could give these more descriptive variables.

gigimscarborough (commented):
Overall some really great code was written here! Nice work, Jimmy! You showed a lot of intention and resourcefulness throughout this project and ultimately got the job done so kudos to you! 🌟

A few things to keep in mind for the future:
- [ ] Being more mindful of our code stylistically i.e. indentation, use of whitespace and new lines.
- [ ] Making sure we are respecting the intention of the problem. i.e making sure if a problem is asking for a new array or object we are returning a new array or object in memory, if a problem is asking you to solve a problem with a loop making sure that loop, not a method is what is actually providing our solution.
- [ ] Internalizing our resources - Being able to find a solution or method on the interwebs is great, but what is more important is that we understand how that code works if we're going to use it. For example a solution that utilizes features of data types we haven't learn yet, or using .indexOf to solve a problem, not realizing that indexOf under the hood in JavaScript is iterative, and we could do the same thing with our loop and reduce our complexity.

Again, Jimmy really nice work! We've come a long way and you're only going further! 🚀

PR #112, opened 2022-01-31
Skills Sharpening Project 

vlopez515:


src/01-functions.js
```diff
@@ -9,7 +9,9 @@
  *  alwaysTrue();
  *  //> true
  */
-function alwaysTrue() {}
+function alwaysTrue() {
+  return true
```
mckiernantim, line 13: Let's get into the habit of adding a  ` ; ` after our return statements as well

src/01-functions.js
```diff
@@ -39,7 +43,9 @@ function greet() {}
  *  add(10, 20);
  *  //> 30
  */
-function add() {}
+function add(a,b) {
```
mckiernantim, line 46: Let's space out our function parameters ` (a,b) ` => ` (a, b) `

src/01-functions.js
```diff
@@ -74,7 +81,9 @@ function multiply() {}
  *  average(10, 6);
  *  //> 8 // 10 + 6 = 16; 16/2 = 8;
  */
-function average() {}
+function average(a,b) {
+  return (a + b)/ 2
```
mckiernantim, line 85: let's make ` (a + b)/2 ` ` (a + b) / 2 `

src/01-functions.js
```diff
@@ -93,7 +102,9 @@ function average() {}
  *  averageThree(10, 11, 19);
  *  //> 10 // 10 + 11 + 19 = 30; 30 / 3 = 10;
  */
-function averageThree() {}
+function averageThree(a,b,c) {
+  return (a + b + c ) / 3
```
mckiernantim, line 106: This is super easy to read!

src/01-functions.js
```diff
@@ -112,7 +123,14 @@ function averageThree() {}
...
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a,b) {
+  if (typeof a === typeof b) {
+    return true
+  } else {
+    return false
+  }
+
```
mckiernantim, line 132: Watch this trailing extra line of space - if whitespace doesn't make our code easier to read we should lose it

src/01-functions.js
```diff
@@ -151,7 +171,10 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
+function joinStringsWithSpaces(a,b,c,d,e) {
+  return a + ' ' + b + ' ' + c + ' ' + d + ' ' + e
```
mckiernantim, line 175: How would we do this with string interpolation?

src/01-functions.js
```diff
@@ -168,7 +191,9 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+  return a.charAt(0) + a.charAt(a.length-1)
```
mckiernantim, line 195: How would we do this with indexing the string ?

src/02-data-types.js
```diff
@@ -95,7 +111,13 @@ function checkIfCharacterIsInString(text, character) {}
  *  isEven(11);
  *  //> false
  */
-function isEven(num) {}
+function isEven(num) {
+  if (num % 2 == 0) {
```
mckiernantim, line 115: We should always aim to use. ` === ` in our comparison.  While we don't have any chance here of a string of zero in this instance its good to know that ` 0 == "0" ` would return true.

src/03-control-flow.js
```diff
@@ -14,7 +14,13 @@
  *  isEqual(10, "10");
  *  //> false
  */
-function isEqual(a, b) {}
+function isEqual(a, b) {
+  if (a === b) {
+  return true
+} else {
```
mckiernantim, line 20: lets indent our ` } else  { `  over one tab. and our ` return ` statements inside our blocks as well

src/03-control-flow.js
```diff
@@ -86,7 +108,13 @@ function positiveNegativeOrZero(a) {}
  *  aroundTheWorldGreeting();
  *  //> "Hello World"
  */
-function aroundTheWorldGreeting(language) {}
+function aroundTheWorldGreeting(language) {
+  if (language == 'german') {
+    return "Hallo Welt"
+  } else if (language == 'spanish') {
+    return "Hola Mundo"
+  } else {return "Hello World"}
```
mckiernantim, line 116: Lets move our block in 116 down one line here

src/03-control-flow.js
```diff
@@ -109,7 +137,18 @@ function aroundTheWorldGreeting(language) {}
  *  aroundTheWorldGreetingWithSwitch();
  *  //> "Hello World"
  */
-function aroundTheWorldGreetingWithSwitch(language) {}
+function aroundTheWorldGreetingWithSwitch(language) {
+  switch (language) {
+    case "german" :
+      return "Hallo Welt"
```
mckiernantim, line 143: As we said in class - the breaks in this case are technically not needed.  I think we should write our ` break ` statements as you have in switch statements.  The bottom line is that not all switch statements will ` return ` and in that case we need to remember to add a ` break `.  Its a good habit to get in to ( and easy to undo if needed );

src/03-control-flow.js
```diff
@@ -128,7 +167,17 @@ function aroundTheWorldGreetingWithSwitch(language) {}
...
+function calculateLetterGrade(grade) {
+  if (grade >= 90) {
+    return 'A'
+  } else if (grade >= 80) {
+    return 'B'
+  } else if (grade >= 70) {
+    return 'C'
+  } else if (grade >= 60) {
+    return 'D'
+  } else {return 'F'}
```
mckiernantim, line 179: This is so easy to read Victor.  The only thing i would say is lets move our final return down one line and add a  ` ; ` after our returns

src/03-control-flow.js
```diff
@@ -163,7 +218,13 @@ function animalCounts(animal, numberOfAnimals) {}
  *  evenOrOdd(48);
  *  //> "Even"
  */
-function evenOrOdd(a) {}
+function evenOrOdd(a) {
+  if (a % 2 == 0) {
+    return "Even"
+  } if (a % 2 !== 0) {
+    return "Odd"
```
mckiernantim, line 225: This is fire.  This is what we want our functions to look like.  Great spacing.  Easy to read.  Add ` ; ` after the returns and you are set.

src/04-arrays.js
```diff
@@ -79,7 +89,10 @@ function getLastElementOfArray(array) {}
  *  addElementToEndOfArray([ 10 ], 9);
  *  //> [ 10, 9 ]
  */
-function addElementToEndOfArray(array, element) {}
+function addElementToEndOfArray(array, element) {
+  array.push(element)
+  return array
```
mckiernantim, line 94: Why is it we can not simply write ` return array.push(element) ` ? Make sure we know why this is - this is!

src/05-objects.js
```diff
@@ -22,7 +24,9 @@ function createEmptyObject() {}
  *  createObjectWithValue(19);
  *  //> { whateverKey: 19 }
  */
-function createObjectWithValue(val) {}
+function createObjectWithValue(val) {
+  return {key: val}
```
mckiernantim, line 28: What happens if with ` return { val } ` ?

src/08-accessing.js
```diff
@@ -148,7 +165,9 @@ function getCityAndState(person) {}
  *  getFullAddress(person);
  *  //> "697 Pine Drive 2A, Staten Island, NY, 10306"
  */
-function getFullAddress(person) {}
+function getFullAddress(person) {
+  return `${person.address.street} ${person.address.unit}, ${person.address.city}, ${person.address.state}, ${person.address.zip}`
```
mckiernantim, line 169: With a large template string like line 169 we should move these variables to their own lines

src/09-word-problems.js
```diff
@@ -80,7 +93,15 @@ function getCartTotal(products) {}
     compareLocations(address1, address2);
     //> "Same city."
  */
-function compareLocations(address1, address2) {}
+function compareLocations(address1, address2) {
+   if (address1.state === address2.state && address1.city !== address2.city) {
+     return "Same state."
```
mckiernantim, line 98: Lets break apart these long chains of logic on their comparison operator
 every ` && ` or ` || ` should be the last item on a line and variable moved one down.

mckiernantim (commented):
Great work Victor we are almost there.  My biggest advice is lets keep our semi-colons consistent and start to really work on our indentation and spacing in function and arithmetic operators.  The sooner we do this the better as good habits are easy to form and bad habits are hard to break - trust me on this.

PR #114, opened 2022-01-31
skills sharpening project

AishaCKleemoff:


AishaCKleemoff:
Last pull for project

src/01-functions.js
```diff
@@ -24,7 +26,9 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet(name) {
+  return "Hello, " + name + ".";
```
mckiernantim, line 30: Can we do this with template literals?  Code looks VERY clean Aisha.  Great job.

src/01-functions.js
```diff
@@ -39,7 +43,9 @@ function greet() {}
  *  add(10, 20);
  *  //> 30
  */
-function add() {}
+function add(a, b) {
+return (a + b);
```
mckiernantim, line 47: Lets indent our return statement over one tab here.  Also we do not need the ` ( ) ` in this case.

src/01-functions.js
```diff
@@ -112,7 +124,14 @@ function averageThree() {}
...
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+  if (typeof a === typeof b){
+    return true;
+  } else {
+    return false;
+  }
+
```
mckiernantim, line 133: Here on 133 we should remove the empty line.  We use whitespace only when we need to separate code based on logic breaks.

src/01-functions.js
```diff
@@ -151,8 +172,9 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
-
+function joinStringsWithSpaces(a, b, c, d, e) {
+  return a + " " + b +  " " + c + " " + d + " " + e;
```
mckiernantim, line 176: This is a good candidate for a template literal

src/02-data-types.js
```diff
@@ -22,7 +24,9 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  return parseInt(val);
```
mckiernantim, line 28: Rememver we can also use ` Number(val) ` in this instance.  ` ParseInt() ` works but will give a different answer if we pass in a decimal like ` 1.5 `;

src/02-data-types.js
```diff
@@ -79,7 +89,9 @@ function convertToWhisperText(text) {}
  *  checkIfCharacterIsInString("hello there", "a");
  *  //> false
  */
-function checkIfCharacterIsInString(text, character) {}
+function checkIfCharacterIsInString(text, character) {
+  return text.includes(character)
```
mckiernantim, line 93: Try to solve this without `includes()` it is a good challenge and will help understand loops.

src/03-control-flow.js
```diff
@@ -28,7 +34,14 @@ function isEqual(a, b) {}
  *  findLarger(19, 7);
  *  //> 19
  */
-function findLarger(a, b) {}
+function findLarger(a, b) {
+ if (a > b){
+   return a;
+ }
+ else {
```
mckiernantim, line 41: Remember in cases like this we actually can remove the ` else ` block in favor of a simple return statement.  If a is not greater than b we can simply ` return  b `

src/03-control-flow.js
```diff
@@ -45,7 +58,20 @@ function findLarger(a, b) {}
...
+  if (a > b) {
+    return a;
+  }
+  else if (b > a){
+  return b;
+}
+  else{
+    return "tie";
+  }
+  }
```
mckiernantim, line 72: Watch your indentation here

src/03-control-flow.js
```diff
@@ -86,7 +122,17 @@ function positiveNegativeOrZero(a) {}
...
+function aroundTheWorldGreeting(language = "english") {
+   if (language.toLowerCase() === "english") {
+     return "Hello World";
+   }
+   else if (language.toLowerCase() ==="german") {
+     return "Hallo Welt";
+   }
+   else if (language.toLowerCase()=== "spanish") {
+     return "Hola Mundo";
+   }
```
mckiernantim, line 134: We need to create a default value here in case no parameter is passed.

src/04-arrays.js
```diff
@@ -135,7 +155,10 @@ function removeElementFromFrontOfArray(array) {}
  *  getMiddleElement([ 10, null, "30" ]);
  *  //> null
  */
-function getMiddleElement(array) {}
+function getMiddleElement(array) {
+  let midEl = Math.floor(array.length / 2);
+  return array [midEl]
```
mckiernantim, line 160: I think we could find a better varaible name -` mid ` or ` middleEl ` or ` midElement ` seems to work a little better.  If we have to abbreviations in one variable name it can start to lose meaning.  Also lets connect ` array ` and `[midEl]`

src/05-objects.js
```diff
@@ -118,7 +141,10 @@ function addKeyToObject(object, key) {}
  *  addKeyValuePairToObject({ up: true }, "left", false);
  *  //> { up: true, left: false }
  */
-function addKeyValuePairToObject(object, key, val) {}
+function addKeyValuePairToObject(object, key, val) {
```
mckiernantim, line 144: All of our previous object functions should have indentation like this one.  It makes it much easier to read.

src/07-even-more-loops.js
```diff
@@ -143,7 +211,16 @@ function findEqual(array, selected) {}
  *  //> [ 1, 11, 2, 3, 4, 9 ]
  */
 
-function removeDuplicates(array) {}
+function removeDuplicates(array) {
+  let newArray = [];
+  array.forEach((duplicate) => {
```
mckiernantim, line 216: Nice work with this ` .forEach() ` call.  However we need to make sure that we understand exactly what this code is doing.  We have not covered higher order functions or callback functions yet.  If you can not explain what exactly this is doing all along the way lets try to solve this with only using loops.

src/07-even-more-loops.js
```diff
@@ -126,7 +185,16 @@ function findLargest(array) {}
...
-function findEqual(array, selected) {}
+function findEqual(array, selected) {
+  for (let i = 0; i < array.length; i++){
+    if (array[i] === selected) {
+      return true;
+    }
+  }
+  return false;
+
+  return   ;
```
mckiernantim, line 196: Line 196 is not needed as we will return out of the function on 191 or 196

src/08-accessing.js
```diff
@@ -32,7 +32,10 @@
  *  getFirstName(person);
  *  //> "Rachel"
  */
-function getFirstName(person) {}
+function getFirstName(person) {
+
```
mckiernantim, line 36: Let's lose this empty line as it doesn't separate any code worth breaking apart.

src/08-accessing.js
```diff
@@ -148,7 +166,9 @@ function getCityAndState(person) {}
  *  getFullAddress(person);
  *  //> "697 Pine Drive 2A, Staten Island, NY, 10306"
  */
-function getFullAddress(person) {}
+function getFullAddress(person) {
+  return `${person.address.street} ${person.address.unit}, ${person.address.city}, ${person.address.state}, ${person.address.zip}`;
```
mckiernantim, line 170: When using backticks we can break strings across multiple lines.  With a long string like this we should spread this out over multiple lines.

src/08-accessing.js
```diff
@@ -171,7 +191,21 @@ function getFullAddress(person) {}
       numbers: [7185550921, 7185558611],
     };
  */
-function getFlatObject(person) {}
+function getFlatObject(person) {
+  let newObj = {};
+  let newObjTwo;
+
+  newObjTwo = Object.assign(newObj, person.names);
```
mckiernantim, line 198: Nice `Object.assign()` call.  Could you explain how this is working step by step?  Also - can you solve this without using ` Object.assign `?

src/09-word-problems.js
```diff
@@ -19,7 +19,25 @@
...
+      total = priceInCents * 0.9;
+    }
+  } else {
+    if (hasMembership === true) {
+      total = priceInCents * 0.8;
+    } else {
+      total = priceInCents;
+    }
+  }
+
```
mckiernantim, line 37: Nice work here.  Let's watch the indentation.  Make sure our logic in the code blocks is indented and we don't have unneeded empty lines.

mckiernantim (commented):
Great start Aisha.  Let's keep an eye on our spacing and indentation to start making good habits early.  Take another pass at the tests and see if we can get them all to pass.

PR #130, opened 2022-01-31
Brandon Mahabir Skill Sharpening Project

BMahabir4112:


src/01-functions.js
```diff
@@ -24,7 +26,10 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet(name) {
+  let greetingMessage = `Hello, ${name}.`
+  return greetingMessage
```
mckiernantim, line 31: If we want we could return the string without saving as a variable

src/01-functions.js
```diff
@@ -112,7 +129,11 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+  if (typeof a == typeof b){
+    return true}
```
mckiernantim, line 134: lets move the ` } ` down to a line.  Also do we need an ` else ` ?

src/01-functions.js
```diff
@@ -131,7 +152,10 @@ function compareTypes() {}
  *  flipSign(50);
  *  //> -50
  */
-function flipSign() {}
+function flipSign(a) {
+oppNancy = a - (2 * a)
```
mckiernantim, line 156: Lets indent a tab here.  Also this is very clever but its there a simpler way?

src/01-functions.js
```diff
@@ -151,7 +175,10 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
+function joinStringsWithSpaces(a, b, c, d, e) {
+let spaceMaster = a + " " + b + " " + c + " " + d + " " + e
```
mckiernantim, line 179: can we do this with template strings?

src/01-functions.js
```diff
@@ -168,7 +195,12 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+  let First = a.charAt(0);
+  let Second = a.charAt(a.length -1);
+  let Boffum = `${First}` + `${Second}`;
```
mckiernantim, line 201: notice the color change in your code?  We do not ever want to Capitalize our variables.  This actually sets up our vars to be something called CLASSES according to JS.  Also - can we solve this without `charAt()` ?

src/02-data-types.js
```diff
@@ -36,7 +40,9 @@ function convertToNumber(val) {}
  *  convertToString(99);
  *  //> "99"
  */
-function convertToString(val) {}
+function convertToString(val) {
+  return val.toString()
```
mckiernantim, line 44: Remember to add a ` ; ` after our returns

src/02-data-types.js
```diff
@@ -95,7 +109,11 @@ function checkIfCharacterIsInString(text, character) {}
  *  isEven(11);
  *  //> false
  */
-function isEven(num) {}
+function isEven(num) {
+  if (num % 2 == 0){
+    return true;}
```
mckiernantim, line 114: lets knock the brace down by one;

src/02-data-types.js
```diff
@@ -127,7 +149,11 @@ function isOdd(num) {}
  *  isTruthy(null);
  *  //> false
  */
-function isTruthy(val) {}
+function isTruthy(val) {
+  if (Boolean(val) == true){
```
mckiernantim, line 153: Can we do this without invoking ` Boolean() ` ?

src/03-control-flow.js
```diff
@@ -14,7 +14,11 @@
  *  isEqual(10, "10");
  *  //> false
  */
-function isEqual(a, b) {}
+function isEqual(a, b) {
+  if (a === b){
+    return true
+  } else return false
```
mckiernantim, line 20: Do we need an  ` else  ` here?

src/03-control-flow.js
```diff
@@ -65,7 +79,14 @@ function findLargerOrTie(a, b) {}
  *  positiveNegativeOrZero(1);
  *  //> "Positive"
  */
-function positiveNegativeOrZero(a) {}
+function positiveNegativeOrZero(a) {
+  if (Math.sign(a) === 1){
+    return "Positive"}
+    else if (Math.sign(a) === (-1)){
```
mckiernantim, line 85: How would we do this without using ` Math.sign() ` ?

src/03-control-flow.js
```diff
@@ -109,7 +138,14 @@ function aroundTheWorldGreeting(language) {}
  *  aroundTheWorldGreetingWithSwitch();
  *  //> "Hello World"
  */
-function aroundTheWorldGreetingWithSwitch(language) {}
+function aroundTheWorldGreetingWithSwitch(language) {
+  switch (language) {
+    case "german": return ("Hallo Welt");
```
mckiernantim, line 143: We should move these ` return `s down below their ` case ` s and indent.  Its easier to see what is returning and where.

src/03-control-flow.js
```diff
@@ -128,7 +164,19 @@ function aroundTheWorldGreetingWithSwitch(language) {}
...
+function calculateLetterGrade(grade) {
+ if (grade >= 90){
+   return "A"}
+   else if ((grade < 90) && (grade >= 80)){
+     return "B"}
+     else if ((grade < 80) && (grade >= 70)){
+      return "C"}
+      else if ((grade < 70) && (grade >= 60)){
+        return "D"} 
+        else return "F"
```
mckiernantim, line 176: Lets line all these up ident wise.  Easier to read

src/03-control-flow.js
```diff
@@ -146,7 +194,11 @@ function calculateLetterGrade(grade) {}
  *  animalCounts("pig", 1);
  *  //> "1 pig"
  */
-function animalCounts(animal, numberOfAnimals) {}
+function animalCounts(animal, numberOfAnimals) {
+  if (numberOfAnimals >= 2){
+    animal += 's'}
```
mckiernantim, line 199: This is really clever!  I like how you change the animal string. Does this change the original string?

src/03-control-flow.js
```diff
@@ -179,7 +236,11 @@ function evenOrOdd(a) {}
  *  evenOrOddWithTernary(8);
  *  //> "Even"
  */
-function evenOrOddWithTernary(a) {}
+function evenOrOddWithTernary(a) {
+  let result =
+    (a % 2 === 0) ? "Even" : "Odd";
```
mckiernantim, line 241: Remember we actually do not need our `  ( ) ` here.  The ternary is only one operation.

src/05-objects.js
```diff
@@ -118,7 +137,12 @@ function addKeyToObject(object, key) {}
  *  addKeyValuePairToObject({ up: true }, "left", false);
  *  //> { up: true, left: false }
  */
-function addKeyValuePairToObject(object, key, val) {}
+function addKeyValuePairToObject(object, key, val) {
+  // let pushObj = {[key]:[val]}
```
mckiernantim, line 141: Lets get into the habit of removing commented out code.  Comments are ok but unused code is not

src/06-loops.js
```diff
@@ -11,8 +11,12 @@
  *  //> 25
  *  (3 + 4 + 5 + 6 + 7 = 25)
  */
-function rangeSum(min, max) {}
-
+function rangeSum(min, max) {
+  let sumOf = {min,max}
```
mckiernantim, line 15: Be careful here on line 15.  We are actually using objet shorthand to create an object that looks like this: ` { min: min, max: max} ` when I think what we want is an array.

src/06-loops.js
```diff
@@ -25,7 +29,11 @@ function rangeSum(min, max) {}
  *  rangeOdd(10, 15);
  *  //> [ 15, 13, 11 ]
  */
-function rangeOdd(min, max) {}
+function rangeOdd(min, max) {
+  for (let j = max; j > min; j--){
```
mckiernantim, line 33: I would suggest iterating forward through an array unless we need to do otherwise

src/06-loops.js
```diff
@@ -61,7 +69,11 @@ function rangeEveryOther(min, max) {}
  *  containsWhileLoop([ "left", "up", "right" ], "down");
  *  //> false
  */
-function containsWhileLoop(array, target) {}
+function containsWhileLoop(array, target) {
+  while (i = array[target]){
```
mckiernantim, line 73: Be careful with our comparison here.  We are DECLARING i to equal the value of array[target] for every iteration.  I do this a lot :)

src/08-accessing.js
```diff
@@ -87,7 +97,9 @@ function getCountOfPhoneNumbers(person) {}
  *  getFirstPhoneNumber(person);
  *  //> 7185550921
  */
-function getFirstPhoneNumber(person) {}
+function getFirstPhoneNumber(person) {
+  return person.numbers.shift()
```
mckiernantim, line 101: Be careful with ` .shift() ` this actually changes the array permanently.  How else can we index the firt array item?

src/08-accessing.js
```diff
@@ -103,7 +115,10 @@ function getFirstPhoneNumber(person) {}
  *  getLastPhoneNumber(person);
  *  //> 7185558611
  */
-function getLastPhoneNumber(person) {}
+function getLastPhoneNumber(person) {
+  let x = person.numbers.shift(-1)
```
mckiernantim, line 119: Take another look at how `.shift()` works under the hood.  What do we call it with?

src/08-accessing.js
```diff
@@ -148,7 +169,10 @@ function getCityAndState(person) {}
  *  getFullAddress(person);
  *  //> "697 Pine Drive 2A, Staten Island, NY, 10306"
  */
-function getFullAddress(person) {}
+function getFullAddress(person) {
+  let fullAddress = person.address.street + ' ' + person.address.unit + ', ' + person.address.city + ', ' + person.address.state + ', ' + person.address.zip
```
mckiernantim, line 173: The better move here is to use a template literal with backticks.  This lets us spread our string across multiple lines of code.  This makes it much easier to read.

src/08-accessing.js
```diff
@@ -171,7 +195,10 @@ function getFullAddress(person) {}
       numbers: [7185550921, 7185558611],
     };
  */
-function getFlatObject(person) {}
+function getFlatObject(person) {
+  let flatObj = {...person.names, ...person.numbers, ...person.address}
```
mckiernantim, line 199: This is very slick but do we know what exactly is happening under the hood here?  What is the name of the ` ... ` operator and what is it doing?

mckiernantim (commented):
Brandon, great start.  Let's watch the spacing on our brackets and our indentation in general.  These are small things but the sooner we can get them as a habit the better.  Take look at the last three files again and see if we can get these to pass.  Also be careful pulling code from online.  Could you explain what  ` ... ` does under the hood?  Do we know what this is called?  Great resourcefulness but we need to know what these do if we submit them in our work.

PR #138, opened 2022-02-05
frozen skills sharpening jb

mckiernantim:


src/01-functions.js
```diff
@@ -24,8 +26,12 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet(name) {
+  name.charAt(0).toUpperCase;
```
mckiernantim, line 30: I like this extra step here to ensure the name is capitalized.  Can we do this without ` charAt() `

src/01-functions.js
```diff
@@ -112,7 +131,17 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+if (typeof a === typeof b) {
+  return true;
+}
+  else {
```
mckiernantim, line 138: Let's watch our indentation and unneeded space after line 136.  Also if we do not ` return true ` on line 136 - do we need an ` else ` at all?

src/01-functions.js
```diff
@@ -151,12 +184,15 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
+function joinStringsWithSpaces(a, b, c, d, e) {
+  let result = a.concat(" ", b, " ", c, " ", d, " ", e);
```
mckiernantim, line 188: can we do this with template strings?

PR #140, opened 2022-02-05
frozen skills sharpening jb

JedeBrahma:


src/01-functions.js
```diff
@@ -24,8 +26,12 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet(name) {
+  name.charAt(0).toUpperCase;
+  return "Hello, " + `${name}` + ".";
```
mckiernantim, line 31: When using template strings we normaly do not use the ` + ` operator to concatenate - template strings are usually used a replacement.  We can write ` Hello, ${name}.`

src/01-functions.js
```diff
@@ -39,7 +45,11 @@ function greet() {}
  *  add(10, 20);
  *  //> 30
  */
-function add() {}
+
+function add(a,b) {
+ let result = a + b;
```
mckiernantim, line 50: Here declaring a variable is not necessary.  This is a fine practice to use now but down the road we should start to think about when we can directly return the result of an operation.  ` return a + b ` is the same functionally.

src/01-functions.js
```diff
@@ -93,7 +108,11 @@ function average() {}
  *  averageThree(10, 11, 19);
  *  //> 10 // 10 + 11 + 19 = 30; 30 / 3 = 10;
  */
-function averageThree() {}
+function averageThree(a, b, c) {
+  let result = (a + b + c) / 3
+  return result;
+
```
mckiernantim, line 114: Let's get in the habit of removing empty whitespace lines if they are not needed or are breaking the logic in a particular spot that is good to break the logic.

src/01-functions.js
```diff
@@ -131,7 +160,11 @@ function compareTypes() {}
  *  flipSign(50);
  *  //> -50
  */
-function flipSign() {}
+function flipSign(a) {
+  let flip = a * (-1);
+  return flip;
```
mckiernantim, line 165: We can also ` return -a `

src/01-functions.js
```diff
@@ -151,12 +184,15 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
+function joinStringsWithSpaces(a, b, c, d, e) {
+  let result = a.concat(" ", b, " ", c, " ", d, " ", e);
```
mckiernantim, line 188: Nice use of `concat()`.  How would we do this with template literals?

src/01-functions.js
```diff
@@ -168,7 +204,11 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+  let aStr = a.charAt(0) + a.charAt(a.length-1);
+  return aStr;
```
mckiernantim, line 209: can we do this with string indexing?

src/02-data-types.js
```diff
@@ -22,7 +25,10 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  result = Math.floor(val);
```
mckiernantim, line 29: Can we do this a ` Number() ` constructor?

src/02-data-types.js
```diff
@@ -79,7 +91,14 @@ function convertToWhisperText(text) {}
  *  checkIfCharacterIsInString("hello there", "a");
  *  //> false
  */
-function checkIfCharacterIsInString(text, character) {}
+function checkIfCharacterIsInString(text, character) {
+  if (text.includes(character)){
```
mckiernantim, line 95: great looking code here Jede.  I'd challenge you to attempt this without ` .includes() `

src/02-data-types.js
```diff
@@ -127,7 +158,23 @@ function isOdd(num) {}
...
+    case undefined:
+      return false;
+    case "":
+      return false;
+    case 0:
+        return false;
+    case false:
+        return false;
+  }
+  return Boolean(val);
```
mckiernantim, line 176: While this works, this is a very verbose way to solve this.  Look up some shorter ways to evaluate how to determine truthiness in JS.

src/03-control-flow.js
```diff
@@ -56,16 +78,30 @@ function findLargerOrTie(a, b) {}
...
+  let result
+  if (a  > 0) {
+    result = "Positive";
+  } else if 
+      (a < 0) {
+      result = "Negative";
+  } else {
+    result = "Zero";
+  }
+
```
mckiernantim, line 100: Let's watch the spacing here.

src/03-control-flow.js
```diff
@@ -86,7 +122,16 @@ function positiveNegativeOrZero(a) {}
...
  *  //> "Hello World"
  */
-function aroundTheWorldGreeting(language) {}
+function aroundTheWorldGreeting(language) {
+  if (language === "spanish"){
+    return "Hola Mundo";
+  } else if (language === "german"){
+    return "Hallo Welt";
+  } else {
+    return "Hello World";
```
mckiernantim, line 131: Nice work  - can we solve this with a default function parameter?

src/03-control-flow.js
```diff
@@ -109,26 +154,57 @@ function aroundTheWorldGreeting(language) {}
...
  *
  * EXAMPLE:
  *  calculateLetterGrade(99);
  *  //> "A"
  */
-function calculateLetterGrade(grade) {}
+function calculateLetterGrade(grade) {
+  if (grade >= 90){
+    result = "A";
+  } else if
```
mckiernantim, line 194: Lets remember to put a code block for our  ` } else if { ` statements.

src/04-arrays.js
```diff
@@ -65,7 +77,10 @@ function getFirstElementOfArray(array) {}
  *  getLastElementOfArray([ null, undefined ]);
  *  //> undefined
  */
-function getLastElementOfArray(array) {}
+function getLastElementOfArray(array) {
+  
+  return array.pop(array.length-1);
```
mckiernantim, line 82: Take another look at what ` .pop() ` does.  What arguments does this function take?

src/04-arrays.js
```diff
@@ -106,7 +130,12 @@ function removeElementFromEndOfArray(array) {}
  *  addElementToFrontOfArray([ 10 ], 9);
  *  //> [ 9, 10 ]
  */
-function addElementToFrontOfArray(array, element) {}
+function addElementToFrontOfArray(array, element) {
+  
+  array.unshift(element);
```
mckiernantim, line 135: Make sure we know why we can't simply ` return array.unshift(element) `

src/05-objects.js
```diff
@@ -22,7 +26,16 @@ function createEmptyObject() {}
  *  createObjectWithValue(19);
  *  //> { whateverKey: 19 }
  */
-function createObjectWithValue(val) {}
+function createObjectWithValue(val) {
+
+  let valObject = {
+
+                  aKey: val
```
mckiernantim, line 33: This object is a little crazy - lets fix our spacing here.

src/05-objects.js
```diff
@@ -35,7 +48,13 @@ function createObjectWithValue(val) {}
  *  createObjectWithKey("left");
  *  //> { left: false }
  */
-function createObjectWithKey(key) {}
+function createObjectWithKey(key) {
+  let keyObject = {}
+                                  
+      keyObject[key] = key;
```
mckiernantim, line 54: same here - spacing

src/05-objects.js
```diff
@@ -118,7 +154,15 @@ function addKeyToObject(object, key) {}
  *  addKeyValuePairToObject({ up: true }, "left", false);
  *  //> { up: true, left: false }
  */
-function addKeyValuePairToObject(object, key, val) {}
+function addKeyValuePairToObject(object, key, val) {
+  newObject = {};
+  newObject[key] = val;
+  //newObject[val] = val;
```
mckiernantim, line 160: We should start getting into the habit of removing and ` console.log() ` statements or commented out code from our files before we push.

src/06-loops.js
```diff
@@ -39,7 +52,15 @@ function rangeOdd(min, max) {}
  *  rangeEveryOther(11, 18);
  *  //> [ 11, 13, 15, 17 ]
  */
-function rangeEveryOther(min, max) {}
+function rangeEveryOther(min, max) {
+  
+  for (let i = min; i <= max; i++) {
```
mckiernantim, line 57: This code works but technically we are still checking every single element of the array.  Can you solve this to only check every other number - in essence taking less steps?

src/06-loops.js
```diff
@@ -39,7 +52,15 @@ function rangeOdd(min, max) {}
...
+    if (i % 2 === 0);
+    array.push(i);
+       
+  }
+  return array.join(", ");
```
mckiernantim, line 62: Take a good look at what ` .join() ` is doing under the hood?

src/06-loops.js
```diff
@@ -61,7 +82,15 @@ function rangeEveryOther(min, max) {}
  *  containsWhileLoop([ "left", "up", "right" ], "down");
  *  //> false
  */
-function containsWhileLoop(array, target) {}
+function containsWhileLoop(array, target) {
+
+  while (array.target = true) {
```
mckiernantim, line 87: what are we doing with line 87 and 90 ?  What does  ` array.target = true ` evaluate to?

src/07-even-more-loops.js
```diff
@@ -12,8 +12,15 @@
  *  shoutForLoop([ "A", "Very", "Happy", "Array" ]);
  *  //> [ "A!", "Very!", "Happy!", "Array!" ];
  */
-function shoutForLoop(array) {}
+function shoutForLoop(array) {
+  
+  for(let element of array) {
+    console.log(`element!`)
 
+   return array;
```
mckiernantim, line 20: Looks like our loops just returns the array on first iteration.  Walk through what exactly is happening on each iteration of your loop to figure out what we need to do to solve.

src/07-even-more-loops.js
```diff
@@ -28,7 +35,16 @@ function shoutForLoop(array) {}
...
-function shoutWhileLoop(array) {}
+function shoutWhileLoop(array) {
+
+  let count = 1;
+  while (count <= array.length) {
+    array.push(count++)
+    console.log(array);
+   }
+   
+   return array["!"];
```
mckiernantim, line 46: line 42 is doing some funny stuff - first it's iterating the value of ` count `.  Second it is pushing that value into your ` array `.  Lin 46 is trying to return the ` array ` at the index of ` ! `.  Remember we can only index arrays by number so 46 will throw an error.  Secondly while iterating through an array we need to determine _what our index_ is and what _the item in the array at the index is_.  These are two crucial concepts to understanding loops

src/08-accessing.js
```diff
@@ -71,7 +80,12 @@ function getAddressStreet(person) {}
  *  getCountOfPhoneNumbers(person);
  *  //> 2
  */
-function getCountOfPhoneNumbers(person) {}
+function getCountOfPhoneNumbers(person) {
+
+  let result = person["numbers"].length;
```
mckiernantim, line 85: remember we can also index ` person.numbers.length `

src/08-accessing.js
```diff
@@ -116,7 +136,13 @@ function getLastPhoneNumber(person) {}
  *  getFullName(person);
  *  //> "Rachel Eleanor Rojas"
  */
-function getFullName(person) {}
+function getFullName(person) {
+  let firstName = person.names.first;
+  let midName = person.names.middle;
+  let lastName = person.names.last;
+  let fullName = firstName.concat(" ", midName ," ", lastName);
```
mckiernantim, line 143: Nice work - I would recommend moving away from the `concat()` method and starting to work with template literals and backticks.  They're just a little smoother to read.

src/08-accessing.js
```diff
@@ -171,7 +212,38 @@ function getFullAddress(person) {}
       numbers: [7185550921, 7185558611],
     };
  */
-function getFlatObject(person) {}
+function getFlatObject(person) {
+ let newPerson = {};
+ newPerson.name = person.name;
```
mckiernantim, line 217: Let's take another look at our assignments here - we have a very good start.  Also I would recommend that if we do nit _specifically have to delete an object key_ not to do it.  Modifying our data permanently is something we can never really come back from so we should avoid it if at all possible.

mckiernantim (commented):
This a good start Jede.  Looks like we are still failing some of our tests in loops, more loops, accessing and word problems.  See if you can make them all pass.  Overall I see some really good habits starting here.  We should really focus in on loops to make sure we have that tricky concept down solid before the end of the module

PR #208, opened 2022-10-17
finished fuonction

touhami-ben:


touhami-ben:
I was waiting for my partner to figure out how to merge our project . so I'm gonna send it as it is

src/08-accessing.js
```diff
@@ -171,7 +189,19 @@ function getFullAddress(person) {}
       numbers: [7185550921, 7185558611],
     };
  */
-function getFlatObject(person) {}
+function getFlatObject(person) {
+  return {
```
Angiepmunoz, line 193: this is not exactly hardcoded but it can be done more dynamically! How can you use a loop to create the flat object?!

src/07-even-more-loops.js
```diff
@@ -96,7 +137,9 @@ function evenArray(array) {}
  *  findSmallest([ 0, 11, -2, 5 ]);
  *  //> -2
  */
-function findSmallest(array) {}
+function findSmallest(array) {
+  return Math.min(...array)
```
Angiepmunoz, line 141: nice use of the spread operator!

PR #210, opened 2022-10-17
isTruthy(), isFalsy()

vg0at:


src/02-data-types.js
```diff
@@ -62,7 +70,9 @@ function convertToShoutingText(text) {}
  *  convertToWhisperText("Hello There");
  *  //> "hello there"
  */
-function convertToWhisperText(text) {}
+function convertToWhisperText(text) {
+  return(text.toLowerCase())
```
Angiepmunoz, line 74: no need to add the extra parenthesis

src/03-control-flow.js
```diff
@@ -86,7 +106,24 @@ function positiveNegativeOrZero(a) {}
...
+  const gerMsg = "Hallo Welt";
+  if (language === 'spanish') {
+    return spanMsg;
+  }
+  else if (language === 'german') {
+    return gerMsg;
+  }
+  else if (language === 'english') {
+    return engMsg;
+  } else if (language !== 'spanish' || language !== 'german' || language !== 'english') {
```
Angiepmunoz, line 121: This is a good place to just use an else. Since you have conditionals that are already accounting for spanish, german, and english, anything else will false under the else.

src/04-arrays.js
```diff
@@ -92,7 +105,9 @@ function addElementToEndOfArray(array, element) {}
  *  removeElementFromEndOfArray([ 10, 9, 8 ]);
  *  //> 8
  */
-function removeElementFromEndOfArray(array) {}
+function removeElementFromEndOfArray(array) {
+return array.pop(array.length-1);
```
Angiepmunoz, line 109: the `pop` method does not take an argument. The proper syntax would be `array.pop()`

src/04-arrays.js
```diff
@@ -106,7 +121,10 @@ function removeElementFromEndOfArray(array) {}
  *  addElementToFrontOfArray([ 10 ], 9);
  *  //> [ 9, 10 ]
  */
-function addElementToFrontOfArray(array, element) {}
+function addElementToFrontOfArray(array, element) {
+  array.unshift(element);
```
Angiepmunoz, line 125: unshift does not take an argument. If you are ever unsure, make sure to take a look at documentation.

src/04-arrays.js
```diff
@@ -119,7 +137,9 @@ function addElementToFrontOfArray(array, element) {}
  *  removeElementFromFrontOfArray([ 10, 9, 8 ]);
  *  //> 10
  */
-function removeElementFromFrontOfArray(array) {}
+function removeElementFromFrontOfArray(array) {
+  return array.shift(array[0]);
```
Angiepmunoz, line 141: shift does not take an argument. Check out the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift).

src/08-accessing.js
```diff
@@ -103,7 +114,9 @@ function getFirstPhoneNumber(person) {}
  *  getLastPhoneNumber(person);
  *  //> 7185558611
  */
-function getLastPhoneNumber(person) {}
+function getLastPhoneNumber(person) {
+  return person.numbers[length-1];
```
Angiepmunoz, line 118: it looks like there some confusion with the length property. When using it you want to make sure you are calling it on whatever `string` or `array` you are attempting to get the length one. 

One way to think about it is that `length-1` does not evaluate to anything. 
What array do you want the length of?  - in this case person.numbers
so `person.numbers.length-1` would evaluate to the last index in the array 
therefore you want `person.numbers[person.numbers.length-1]`

src/08-accessing.js
```diff
@@ -171,7 +191,21 @@ function getFullAddress(person) {}
       numbers: [7185550921, 7185558611],
     };
  */
-function getFlatObject(person) {}
+function getFlatObject(person) {
+  const flatObj = {
```
Angiepmunoz, line 195: this is hardcoded. How can you use a loop to make it more dynamic??

PR #227, opened 2022-10-21
Riya Skill sharpening w/ Christian O. and Ari C.

riyaalexander:


src/08-accessing.js
```diff
@@ -171,7 +192,21 @@ function getFullAddress(person) {}
       numbers: [7185550921, 7185558611],
     };
  */
-function getFlatObject(person) {}
+function getFlatObject(person) {
```
Angiepmunoz, line 195: This is not exactly hardcoded but it can definitely be refactored to be more dynamic. How can you use a loop to create the flat object?

PR #228, opened 2022-10-21
Project

joshuanelsondev:
In a group with Elizabeth Bright and Taylor Brooks.

### DIFF
--- a/src/01-functions.js
+++ b/src/01-functions.js
@@ -54,7 +54,10 @@ function add() {}
  *  multiply(2, 5);
  *  //> 10
  */
-function multiply() {}
+function multiply(a, b) {
+  
+  return a * b;
+}
 
 /**
  * average()
@@ -74,7 +77,10 @@ function multiply() {}
  *  average(10, 6);
  *  //> 8 // 10 + 6 = 16; 16/2 = 8;
  */
-function average() {}
+function average(a, b) {
+
+  return (a + b) / 2;
+}
 
 /**
  * averageThree()
@@ -93,7 +99,10 @@ function average() {}
  *  averageThree(10, 11, 19);
  *  //> 10 // 10 + 11 + 19 = 30; 30 / 3 = 10;
  */
-function averageThree() {}
+function averageThree(a, b, c) {
+
+  return (a + b + c) / 3;
+} 
 
 /**
  * compareTypes()
--- a/src/01-functions.js
+++ b/src/01-functions.js
@@ -9,7 +9,9 @@
  *  alwaysTrue();
  *  //> true
  */
-function alwaysTrue() {}
+function alwaysTrue() {
+  
+}
 
 /**
  * greet()
@@ -24,7 +26,9 @@ function alwaysTrue() {}
  *  greet("Ian");
  *  //> "Hello, Ian."
  */
-function greet() {}
+function greet() {
+
+}
 
 /**
  * add()
@@ -112,7 +116,10 @@ function averageThree() {}
  *  compareTypes("left", 10);
  *  //> false
  */
-function compareTypes() {}
+function compareTypes(a, b) {
+
+  return typeof a === typeof b;
+}
 
 /**
  * flipSign()
@@ -131,7 +138,9 @@ function compareTypes() {}
  *  flipSign(50);
  *  //> -50
  */
-function flipSign() {}
+function flipSign(a) {
+    return a * -1
+}
 
 /**
  * joinStringsWithSpaces()
@@ -151,7 +160,10 @@ function flipSign() {}
  *  joinStringsWithSpaces("left", "right", "up", "down", "away");
  *  //> "left right up down away"
  */
-function joinStringsWithSpaces() {}
+function joinStringsWithSpaces(a, b, c, d, e) {
+
+  return `${a} ${b} ${c} ${d} ${e}`;
+}
 
 /**
  * getFirstAndLastCharacter()
@@ -168,7 +180,10 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+
+  return `${a[0]}${a[a.length - 1]}`;
+}
 
 // Do not change the code below.
 module.exports = {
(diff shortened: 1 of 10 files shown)


src/01-functions.js
```diff
@@ -168,7 +191,10 @@ function joinStringsWithSpaces() {}
  *  getFirstAndLastCharacter("upwards");
  *  //> "us"
  */
-function getFirstAndLastCharacter() {}
+function getFirstAndLastCharacter(a) {
+
+  return `${a[0]}${a[a.length - 1]}`;
```
abbreviatedman: Using template literals when most or all of the string is evaluated JS can be tough to read. You might try `a[0] + a[a.length - 1]` for clearer code.

src/02-data-types.js
```diff
@@ -22,7 +24,9 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  return Number(val);
```
abbreviatedman: Great job here! Using `Number` (rather than `parseFloat` or `parseInt` or `+`) adds a lot to the code's readability.

src/02-data-types.js
```diff
@@ -79,7 +89,9 @@ function convertToWhisperText(text) {}
  *  checkIfCharacterIsInString("hello there", "a");
  *  //> false
  */
-function checkIfCharacterIsInString(text, character) {}
+function checkIfCharacterIsInString(text, character) {
+  return text.includes(character);
```
abbreviatedman: If your value is _already_ the `true` or `false` you need, you can simply return that value, as you did here with `.includes`. There's no need to tack on a `? true : false`, and you wisely saw that you could skip that.

See the solution below where you use an unnecessary ternary—`num % 2 === 0`, for example, is already the boolean you want. Apply what you did here to those problems!

src/03-control-flow.js
```diff
@@ -146,7 +195,9 @@ function calculateLetterGrade(grade) {}
  *  animalCounts("pig", 1);
  *  //> "1 pig"
  */
-function animalCounts(animal, numberOfAnimals) {}
+function animalCounts(animal, numberOfAnimals) {
+  return numberOfAnimals > 1 ? `${numberOfAnimals} ${animal}s` : `1 ${animal}`;
```
abbreviatedman: Although this solution passes the test, you should extend it to one more case: what would happen if there were 0 animals? Re-configuring your solution a bit will solve that easily!

src/04-arrays.js
```diff
@@ -24,7 +27,9 @@ function createEmptyArray() {}
  *  createArrayWithTwoElements(true, false);
  *  //> [ true, false ]
  */
-function createArrayWithTwoElements(a, b) {}
+function createArrayWithTwoElements(a, b) {
+  return [a, b];
```
abbreviatedman: Great job returning the array literal, rather than going through extra steps! Very compact solution.

src/04-arrays.js
```diff
@@ -135,7 +155,9 @@ function removeElementFromFrontOfArray(array) {}
  *  getMiddleElement([ 10, null, "30" ]);
  *  //> null
  */
-function getMiddleElement(array) {}
+function getMiddleElement(array) {
+  return array[(array.length - 1) / 2];
+}
```
abbreviatedman: To extend this, think about what would happen if you had an even nunber of elements in the array.

src/05-objects.js
```diff
@@ -35,7 +40,11 @@ function createObjectWithValue(val) {}
  *  createObjectWithKey("left");
  *  //> { left: false }
  */
-function createObjectWithKey(key) {}
+function createObjectWithKey(key) {
+  let obj = {}
+  obj[key] = "anything"
+  return obj;
```
abbreviatedman: You can return an object literal here and make it one line. Start with `return {}` and see if you can fill in what could go into those curly brackets!

src/06-loops.js
```diff
@@ -25,7 +31,16 @@ function rangeSum(min, max) {}
...
+function rangeOdd(min, max) {
+  let sum = [];
+  for (let i = max; i >= min; i--) {
+    if (i % 2 !== 0){
+      sum.push(i);
+    }
+  } 
+  
+  return sum;
+}
```
abbreviatedman: You folks are solving problems to the point where it's time to think beyond getting the problem correct. In this case, I'd encourage you to think for a moment as you name variables—is this name as specific and accurate as it can be? In this case, your return value is not a sum!

src/07-even-more-loops.js
```diff
@@ -44,7 +61,15 @@ function shoutWhileLoop(array) {}
...
-function shoutForOfLoop(array) {}
+function shoutForOfLoop(array) {
+  const arr3 = []
+ for (let ele of array){
+  arr3.push(ele + "!");
+ }
+
+
+  return arr3;
+}
```
abbreviatedman: Keep scope in mind—there is no need for an `arr3` name here, as there's no `arr2` in this scope.

You might also give them different names based on what they are. `arr3` could be `appendedStrings` or `shouts` or `result` or `answer`. Whatever name makes it so that you never have to think, "What is in that array, again?"

src/07-even-more-loops.js
```diff
@@ -96,7 +147,16 @@ function evenArray(array) {}
...
+function findSmallest(array) {
+  let smallNum = array[0];
+  for (let i = 0; i < array.length; i++) {
+    if (array[i] < smallNum) {
+      smallNum = array[i];
+    }
+  }
+
+  return smallNum;
+}
```
abbreviatedman: Excellent accumulation!

src/09-word-problems.js
```diff
@@ -127,7 +159,30 @@ function compareLocations(address1, address2) {}
...
+      }
+    } else {
+        assignments[i].status = `SCORE: ${assignments[i].score.received}/${assignments[i].score.max}`;
+    }  
+      
+  } 
+  
+      return assignments;
+
+}
```
abbreviatedman: Nested `if/else` chains are easy to have a bug in and difficult to update. Stretch yourself by making the main `if/else-if` chain into a `switch` statement!

src/09-word-problems.js
```diff
@@ -152,7 +208,22 @@ function gradeAssignments(assignments) {}
...
+    } else {
+      arr2.push(people[i].name)
+    } 
+
+  }
+
+
+
+  return arr.concat(arr2);
+}
```
abbreviatedman: Great solution, here!

PR #234, opened 2022-10-21
Skill Sharpening Project Jordan Marangely Christine

Jordan-Walker-Web:


src/01-functions.js
```diff
@@ -9,7 +9,14 @@
  *  alwaysTrue();
  *  //> true
  */
-function alwaysTrue() {}
+function alwaysTrue() {
+  let a = true
+  return a;
```
Angiepmunoz, line 14: another way to do it, is to simply return `true` instead of storing it in a variable. Either is correct tho!

src/01-functions.js
```diff
@@ -93,7 +118,12 @@ function average() {}
  *  averageThree(10, 11, 19);
  *  //> 10 // 10 + 11 + 19 = 30; 30 / 3 = 10;
  */
-function averageThree() {}
+function averageThree(a, b, c) {
+
+  let average3 = (a + b + c) /3;
+
+  return average3;
```
Angiepmunoz, line 125: Same goes for this one and the few above it

src/03-control-flow.js
```diff
@@ -86,7 +138,26 @@ function positiveNegativeOrZero(a) {}
...
+}
+  else if (language === "spanish") {
+  return "Hola Mundo"; 
+  } 
+  else if (language === "german") {
+  return "Hallo Welt";
+  
+} else {
+
+return "Hello World";
```
Angiepmunoz, line 154: Be careful with formatting. While it doesn't impact your code it's good practice to structure your code with the proper indentation.

src/03-control-flow.js
```diff
@@ -109,7 +180,25 @@ function aroundTheWorldGreeting(language) {}
  *  aroundTheWorldGreetingWithSwitch();
  *  //> "Hello World"
  */
-function aroundTheWorldGreetingWithSwitch(language) {}
+function aroundTheWorldGreetingWithSwitch(language) {
+ switch (language) {
+  case 'english':
+    return 'Hello World';
+    break;
```
Angiepmunoz, line 187: Since the `return` stops your code from running, you don't need to include a `break`. The returns acts like the `break`

src/04-arrays.js
```diff
@@ -92,7 +108,11 @@ function addElementToEndOfArray(array, element) {}
  *  removeElementFromEndOfArray([ 10, 9, 8 ]);
  *  //> 8
  */
-function removeElementFromEndOfArray(array) {}
+function removeElementFromEndOfArray(array) {
+      const a =  array.pop(array);
```
Angiepmunoz, line 112: pop does not take an argument. Check out the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop) for the proper syntax.

src/07-even-more-loops.js
```diff
@@ -57,7 +81,17 @@ function shoutForOfLoop(array) {}
...
  *  //> 31
  */
-function sumArray(array) {}
+function sumArray(array) {
+let sum = 0;
+let newArr = [];
+
+for (let i = 0; i < array.length; i++) {
+   newArr.push(array[i])
+sum = newArr.reduce((a, b) => a + b);
```
Angiepmunoz, line 90: In this function, if you use reduce you shouldn't use a loop. Since reduce is lopping through the array, you can apply the method directly tot the original `array` and it will give you the same outcome. 

Have reduce inside the loop, is almost like a nested loop and is running on every iteration.

src/07-even-more-loops.js
```diff
@@ -143,7 +223,13 @@ function findEqual(array, selected) {}
  *  //> [ 1, 11, 2, 3, 4, 9 ]
  */
 
-function removeDuplicates(array) {}
+function removeDuplicates(array) {
+   return array.filter((a,b) => array.indexOf(a) === b)
```
Angiepmunoz, line 227: nice use of filter!

src/08-accessing.js
```diff
@@ -171,7 +193,22 @@ function getFullAddress(person) {}
...
+      last: `${person.names.last}`,
+      street: `${person.address.street}`,
+      unit: `${person.address.unit}`,
+      city: `${person.address.city}`,
+      state: `${person.address.state}`,
+      zip: `${person.address.zip}`,
+      numbers: person.numbers
+  }
+   return newPerson
+  }
```
Angiepmunoz, line 210: while this is not exactly hardcoded, there is a more dynamic way to do it! How can you use a loop to get the flat obj?

PR #237, opened 2022-10-21
This is my code for sharping skill for group

CortezFernando:


src/01-functions.js
```diff
@@ -74,7 +84,12 @@ function multiply() {}
  *  average(10, 6);
  *  //> 8 // 10 + 6 = 16; 16/2 = 8;
  */
-function average() {}
+function average(a, b) {
+  //const result = [ ];
+  const arr = [a, b];
+  const average = arr.reduce((a, b) => a + b, 0) / arr.length;
+  return average
+}
```
abbreviatedman, line 92: Since this is only two numbers being averaged, `.reduce` is overkill. Try using addition and division only for clearer code, as you did with the below average-of-three function.

src/01-functions.js
```diff
@@ -109,10 +127,17 @@ function averageThree() {}
...
  */
-function compareTypes() {}
+function compareTypes(a, b)
+ {
+  if (typeof a == typeof b)  
+  return true
+  else (typeof a != typeof b) 
+   
+ return false
+}
```
abbreviatedman, line 140: Since `typeof a == typeof b` returns a boolean that will be `true` when you want to return `true` and `false` when you want to return `false`, you can simply return the value it evaluates to.

The equivalent with numbers would be this function:

``` javascript
function add(a, b) {
  const sum = a + b;
  if (sum === 1) {
    return 1;
  } else if (sum === 2) {
    return 2;
  } else if (sum === 3) {
    return 3;
  } else if (sum === 4) {
    return 4;
  } else // you get the idea
}
```

You could simply return `a + b` in this case, and you can simply return the `typeof` calculation in the same way. Try it!

`

src/02-data-types.js
```diff
@@ -22,7 +24,9 @@ function getLengthOfString(str) {}
  *  convertToNumber("111");
  *  //> 111
  */
-function convertToNumber(val) {}
+function convertToNumber(val) {
+  return +val
+}
```
abbreviatedman, line 29: `Number(val)` would be clearer, but others prefer `parseInt` or `parseFloat`. `+` works too, but clarity is king!

src/02-data-types.js
```diff
@@ -79,8 +89,28 @@ function convertToWhisperText(text) {}
...
  */
-function checkIfCharacterIsInString(text, character) {}
+function checkIfCharacterIsInString(text, character) {
+  if (text.includes(character)) {
+    return true
+  }
+  else {
+    return false
+  }
+}
```
abbreviatedman, line 99: Same point as with `compareTypes`. `.includes` already gives you the value you want to return, no need to check its value before returning it.

src/02-data-types.js
```diff
@@ -95,7 +125,13 @@ function checkIfCharacterIsInString(text, character) {}
...
  *  //> false
  */
-function isEven(num) {}
+function isEven(num) {
+  if (num % 2 === 0)
+    return true
+  else {
+    return false
+  }
+}
```
abbreviatedman, line 134: Same point as `compareTypes`.

src/03-control-flow.js
```diff
@@ -14,7 +14,9 @@
  *  isEqual(10, "10");
  *  //> false
  */
-function isEqual(a, b) {}
+function isEqual(a, b) {
+  return a === b
+}
```
abbreviatedman, line 19: Just what I'm talking about above. Well done!

src/03-control-flow.js
```diff
@@ -45,27 +49,48 @@ function findLarger(a, b) {}
  *  findLargerOrTie(0, 0);
  *  //> "tie"
  */
-function findLargerOrTie(a, b) {}
+function findLargerOrTie(a, b) {
+  return Math.max(a, b) | "tie"
+
+}
```
abbreviatedman: Would be clearer with an `if` check for equality, I think. Can you explain the `|` operator? Awesome if you've got that down, but I wouldn't use it if you don't!

src/04-arrays.js
```diff
@@ -65,7 +74,13 @@ function getFirstElementOfArray(array) {}
  *  getLastElementOfArray([ null, undefined ]);
  *  //> undefined
  */
-function getLastElementOfArray(array) {}
+function getLastElementOfArray(array) {
+
+  return array.pop();
+}
```
abbreviatedman, line 80: This is problematic in that it actually removes the last element of the array, which would be an unexpected result when you're asked to "Return" the element, not "Remove". Could introduce some serious bugs!

I would try to rewrite this one without `pop`. How can you access the last index of the array? The JSDoc has a hint!

src/04-arrays.js
```diff
@@ -79,7 +94,9 @@ function getLastElementOfArray(array) {}
  *  addElementToEndOfArray([ 10 ], 9);
  *  //> [ 10, 9 ]
  */
-function addElementToEndOfArray(array, element) {}
+function addElementToEndOfArray(array, element) {
+  return [array, element]
+}
```
abbreviatedman, line 99: This will pass the test, but the array is not what it ought to be. If I call `addElementToEndOfArray` with `[3, 5, 6]` and `4`, instead of `[3, 5, 6, 4]` I will get `[[3, 5, 6], 4]` an array with two elements, the first of which is the original array. Further, the `array` hasn't been changed, but a new one returned.

There is an array method that will solve this for you!

src/04-arrays.js
```diff
@@ -92,7 +109,9 @@ function addElementToEndOfArray(array, element) {}
  *  removeElementFromEndOfArray([ 10, 9, 8 ]);
  *  //> 8
  */
-function removeElementFromEndOfArray(array) {}
+function removeElementFromEndOfArray(array) {
+  return array.pop(-1)
+}
```
abbreviatedman, line 114: Check the documentation for `pop`. How many arguments does it take?

src/04-arrays.js
```diff
@@ -106,7 +125,9 @@ function removeElementFromEndOfArray(array) {}
  *  addElementToFrontOfArray([ 10 ], 9);
  *  //> [ 9, 10 ]
  */
-function addElementToFrontOfArray(array, element) {}
+function addElementToFrontOfArray(array, element) {
+  return [element, array]
+}
```
abbreviatedman, line 130: Same comment as for `addElementToEndOfArray`.

src/05-objects.js
```diff
@@ -49,8 +58,13 @@ function createObjectWithKey(key) {}
  *  createObjectWithKeyValuePair("left", "right");
  *  //> { left: "right" }
  */
-function createObjectWithKeyValuePair(key, val) {}
-
+function createObjectWithKeyValuePair(key, val) {
+  return Object.assign({ [key]: val })
+}
```
abbreviatedman, line 63: No need for `Object.assign` here! Try solving without it.

src/05-objects.js
```diff
@@ -103,7 +124,9 @@ function addValueToObject(object, val) {}
  *  addKeyToObject({ right: false }, "left");
  *  //> { right: false, left: true }
  */
-function addKeyToObject(object, key) {}
+function addKeyToObject(object, key) {
+  return Object.assign(object, { [key] : "my key" })
+}
```
abbreviatedman, line 129: Don't need Object.assign here!

src/05-objects.js
```diff
@@ -118,7 +141,9 @@ function addKeyToObject(object, key) {}
  *  addKeyValuePairToObject({ up: true }, "left", false);
  *  //> { up: true, left: false }
  */
-function addKeyValuePairToObject(object, key, val) {}
+function addKeyValuePairToObject(object, key, val) {
+  return Object.assign(object, { [key] : val })
+}
```
abbreviatedman, line 146: Don't need `Object.assign` here!

src/06-loops.js
```diff
@@ -9,10 +9,56 @@
...
+
+//let sum = 0;
+// for (let i = 0; i < arr.length; i++) {
+//   sum += Math.max(arr[i]) + Math.min(arr[i]);
+
+// }
+// return sum 
+
+//pass the test but not a loop =>return (max - min + 1) * (min + max) / 2
+//how many # there are^____how many in the middle^
```
abbreviatedman, line 58: You're returning the starting value. What is it you're being asked to return?

See your solution to `rangeOdd` below!

src/06-loops.js
```diff
@@ -39,7 +94,24 @@ function rangeOdd(min, max) {}
...
+  }
+  else {
+    for (let i = min; i <= max; i++) {
+      if (i % 2 != 0) {
+        arr.push(i) }
+    } 
+  }
+return arr
+}
+
```
abbreviatedman, line 111: Neat solution! Could you do it with 1 loop? Try changing the three parts of your `for` loop and see what happens.

src/07-even-more-loops.js
```diff
@@ -44,7 +63,26 @@ function shoutWhileLoop(array) {}
...
+}
+//Brainstorming & Notes ---V
+//claims its not a for loop in the npm test result->needs a for...of loop
+// for (let value of i = 0) {
+//   i < array.length
+
+//   newArr.push(array[i] + '!')
+//   i++;
+//}
+//element is i in array, what we are looking through the array for an element
```
abbreviatedman, line 85: This works,  but you're using `i` when you don't have to. This is the beauty of the `for of` loop. Try solving this one without `i`!

src/06-loops.js
```diff
@@ -61,7 +133,18 @@ function rangeEveryOther(min, max) {}
...
+  let i = 0
+    while (i < array.length) {
+      if (array[i] === target) {
+        return true
+      }
+      i++
+    }
+    return false
+}
+
```
abbreviatedman, line 144: Great job on the `contains` functions. Returning early as soon as you have an answer is a pro move, right there!

src/07-even-more-loops.js
```diff
@@ -70,8 +115,26 @@ function sumArray(array) {}
...
+
+
+}
+//array.filter(n => n % 2)
+//
+//Brainstorming-->
+//const odds = array.filter(number => 2)
+// {
+//   return number % 2 !== 0;
+// }
```
abbreviatedman, line 137: Great job!

src/07-even-more-loops.js
```diff
@@ -96,7 +176,17 @@ function evenArray(array) {}
...
+function findSmallest(array) {
+  let smallest = array[0]
+  for (i =0; i < array.length; i++){
+    if (array[i] <= smallest){
+      smallest = array[i]
+    }
+  }
+return smallest
+}
+
```
abbreviatedman, line 189: Excellent use of the accumulator!

src/07-even-more-loops.js
```diff
@@ -143,7 +252,18 @@ function findEqual(array, selected) {}
...
+  let newArr = []
+  for (let i = 0; i <array.length; i++){
+    if (newArr.includes(array[i])=== false) {
+      newArr.push(array[i])
+    }
+  }
+  return newArr 
+
+}
+
```
abbreviatedman, line 266: Great solution! We'll explore a more efficient solution when we talk algorithms down the road.

src/08-accessing.js
```diff
@@ -71,7 +77,10 @@ function getAddressStreet(person) {}
  *  getCountOfPhoneNumbers(person);
  *  //> 2
  */
-function getCountOfPhoneNumbers(person) {}
+function getCountOfPhoneNumbers(person) {
+
+  return person.numbers.push()
+}
```
abbreviatedman, line 83: I believe this is one place where you are changing the object passed in to you, as we discussed. There's an array property that can tell you how many are in the array without _changing_ the array.

src/08-accessing.js
```diff
@@ -103,7 +114,9 @@ function getFirstPhoneNumber(person) {}
  *  getLastPhoneNumber(person);
  *  //> 7185558611
  */
-function getLastPhoneNumber(person) {}
+function getLastPhoneNumber(person) {
+  return person.numbers.pop()
+}
```
abbreviatedman, line 119: This is another method that changes the array (and the object it's on). This is the source of your bug, I believe!

src/08-accessing.js
```diff
@@ -171,7 +202,30 @@ function getFullAddress(person) {}
...
+  console.log(person.numbers)
+  const newObj = {}
+  let names = person.names
+  let address = person.address
+
+  Object.assign(newObj, names, address)
+  newObj.numbers = JSON.parse(JSON.stringify(person.numbers))
+  console.log(newObj.numbers)
+  return newObj
+}
```
abbreviatedman, line 227: Check the above comments for why your `numbers` array has been changed!

src/09-word-problems.js
```diff
@@ -19,7 +19,58 @@
...
+  } else if (age <= 10 || age >= 65 && priceInCents * 0.10) {
+    return priceInCents
+  } else if (hasMembership === true && priceInCents * 0.20) {
+    return priceInCents
+  } else if (hasMembership && age <= 10 || age >= 65 || priceInCents * 0.30) {
+    return priceInCents
+  } else {
+    return priceInCents
+  }
+}
```
abbreviatedman, line 41: If you run the debugger on this problem using the input from the first failing test, you'll find some surprising things about which `if` conditions are evaluating to `true` and how the value of `priceInCents` changes before you return it.

PR #264, opened 2022-10-24
Skill Sharpening Project by Diandre, Jennifer and Vandhana

Vandhana-Mohan:


src/06-loops.js
```diff
@@ -39,7 +54,18 @@ function rangeOdd(min, max) {}
...
  *  //> [ 11, 13, 15, 17 ]
  */
-function rangeEveryOther(min, max) {}
+
+/********************************* VANDHANA MOHAN *******************************/
+
+
+function rangeEveryOther(min, max) { // return odd numbers
+  let newrangeEveryOther = []
+  for(let i = min; i <= max; i+=2){
```
Angiepmunoz, line 63: nice job using the reassignment operator to increase by 2!

src/08-accessing.js
```diff
@@ -171,7 +211,21 @@ function getFullAddress(person) {}
       numbers: [7185550921, 7185558611],
     };
  */
-function getFlatObject(person) {}
+function getFlatObject(person) {
+  const Diandre = {
```
Angiepmunoz, line 215: This isn't entirely hardcoded but there is a way more dynamic way to create the flat object. How can you use a for in loop to dynamically create the flat object?

